PR #1, opened 2024-06-07
feat: user activity watch

gentlementlegen:
Resolves https://github.com/ubiquibot/plugins-wishlist/issues/7


gentlementlegen:
QA run: https://github.com/Meniole/user-activity-watcher/issues/2#issuecomment-2157278506 (timers set to 1 minute for inactivity, 5 minutes for un-assignment)

src/handlers/issue-closed.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * On issue closed, we want to delete the entry to stop watching this issue.
+ */
+export async function handleIssueClosed(context: Context, env: EnvConfigType): Promise<Result> {
+  const {
+    adapters: {
+      supabase: { repositories },
+    },
+  } = context;
+  await repositories.delete(context.payload.issue.html_url);
```
0x4007, line 14: You just reminded me that GitHub has some type of feature to associate metadata with each repository. Perhaps we can write whatever data we need here for minimizing runs?

> ## Custom Properties for Repositories
> GitHub also allows organizations to manage custom properties for their repositories. These properties can be used to add metadata such as compliance frameworks, data sensitivity, or project details. Custom properties can be viewed by people with read permissions to the repository and can be used to search and filter repositories within an organization.
> ### Adding Custom Properties
> Navigate to the repository's settings.
> In the "Code and automation" section, click "Custom properties".
> Click "New property" to add a new custom property.
> Fill in the name, description, and type of the property.
> Optionally, set permissions and default values.
> Save the property.
> ### Setting Values for Custom Properties
> Navigate to the organization's settings.
> In the "Code, planning, and automation" section, click "Repository" and then "Custom properties".
> Select the repositories and click "Edit properties".
> Set the values for each property and save the changes.
> These methods provide a comprehensive way to access and manage GitHub repository metadata, whether through automated workflows, API calls, or custom properties within an organization.

rndquu:
@gentlementlegen I'm getting `[ERROR] SyntaxError: "[object Object]" is not valid JSON` error when running this plugin as a worker with [this](https://github.com/rndquu-org/test-repo/blob/bb82387db22e7f8c46a024c20a82605d000af935/.github/.ubiquibot-config.yml) config (mostly taken from README). What am I doing wrong?

gentlementlegen:
@rndquu It is meant to be an Action not a Worker, at first I tried to make it as a worker but I realized it was a long running / recurring task, will remove unnecessary files apologies. Please use it as an Action instead.

rndquu:
@gentlementlegen Is [this one](https://github.com/rndquu-org/test-repo/blob/9a57f04eae649d31b7564541a2408055ef17bafe/.github/.ubiquibot-config.yml) a valid syntax for calling a specific branch of the plugin? I took this example from README at https://github.com/ubiquibot/plugin-template.

gentlementlegen:
@rndquu should be accepting the same syntax as Github although your example is very specific and not sure we took this into account. You can specify workflow and branch usually. If this is not working and this is a valid Github workflow, please file an issue for it

rndquu:
@gentlementlegen Help

I'm getting the `Failed to run user-activity-watcher: Error: [parseGitHubUrl] Invalid url: [https://github.com/rndquu-org/test-repo]` error in [this](https://github.com/rndquu/user-activity-watcher/actions/runs/9455000675/job/26043811384) workflow. What URL is expected to be in the `repositories.url` column in the Supabase DB?

P.S. It seems that RLS policies are not applied in the [migration script](https://github.com/ubiquibot/user-activity-watcher/pull/1/files#diff-27c68f9d4869303fb1fd77f691d6b7cfe1a7a5444ce411c75a91b07280803465R11). So they should be added manually via Supabase UI to enable them and to allow read access for public anon key.

gentlementlegen:
@rndquu The expected URL should point to the issue itself, because the issue number is used to retrieve information about the status of it, e.g. https://github.com/rndquu-org/test-repo/issues/56

Also you are right, I just enabled them but didn't set any specific rule and used the elevated privileges key. I can add them, but I should give it a read-only level, which will be problematic when it comes to updates the column values.

rndquu:
@gentlementlegen How github issues (that should be watched for activities in the `repositories.url` DB field) are created in the first place? I mean this plugin (as far as I understand) read already saved github issues from a DB and check for assignee activities but doesn't insert any github issues itself. So there should be a piece of code that initially inserts github issues that should be watched.

supabase/migrations/20240603110253_init.sql
```diff
@@ -0,0 +1,63 @@
...
+
+CREATE UNIQUE INDEX repositories_pkey ON public.repositories USING btree (id);
+
+CREATE UNIQUE INDEX repositories_url_key ON public.repositories USING btree (url);
+
+alter table "public"."repositories" add constraint "repositories_pkey" PRIMARY KEY using index "repositories_pkey";
+
+alter table "public"."repositories" add constraint "repositories_url_key" UNIQUE using index "repositories_url_key";
+
+grant delete on table "public"."repositories" to "anon";
```
rndquu, line 21: @gentlementlegen Why do we need to grant so many sensitive (delete, insert, update, etc...) permissions to the `anon` users?

gentlementlegen:
@rndquu Whenever a `issues.assigned` event is received by the plugin, a new entry will be saved to watch that issue (so inserted in the DB).
If a `issues.unassigned` or `issues.closed` event is the instigator, then that issue will be removed from the watching list (removed from the DB).
Any other event would make the plugin loop through the watched list and update data accordingly.

The main issue with the old bot is that if there is no event on the issue itself, users never get unassigned or reminded (I actually never ever saw a single reminder being sent). This is because updates would only occur on event in the said issue. Keeping track of them allows the plugin to actually refresh all the watched issues on any given event within the whole organization, so we are sure that all the issues where there is an assigned are updated. Second common scenario with the old bot in when the issues stales for some time and then we close the related PR successfully, the user gets unassigned during the incentive evaluation, which I also aimed to solve.
Third reason is that it allows to keep track of last checked time and last checked reminder, otherwise the plugin would have to parse the whole issue with comments every run.

In an ideal world, we should be able to schedule tasks for running at specific times which would simplify the whole logic greatly. But Actions do not support scheduling except CRON, scheduling within Cloudflare requires a paid plan, and we do not have such feature within the Kernel..

rndquu:
@gentlementlegen So what is the expected config that partners should use? This one:
```
plugins:
  issues.assigned:
    - uses:
        - plugin: ubiquibot/user-activity-watcher
          type: github
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
  issues.unassigned:
    - uses:
        - plugin: ubiquibot/user-activity-watcher
          type: github
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
  issues.closed:
    - uses:
        - plugin: ubiquibot/user-activity-watcher
          type: github
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
```

gentlementlegen:
```
plugins:
  '*':
    - uses:
        - plugin: ubiquibot/user-activity-watcher
          type: github
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
```
would be preferred to ensure refresh more often.

rndquu:
@gentlementlegen Does the kernel support this syntax?
```
plugins:
  issues.assigned:
  issues.unassigned:
  issues.closed:
    - uses:
        - plugin: ubiquibot/user-activity-watcher
          type: github
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
```

If not, should we support it? At least it could be useful for `ubiquibot/user-activity-watcher`.

src/proxy/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface Result {
+  status: "ok" | "failed" | "skipped";
+  content?: string;
+  reason?: string;
+}
+
+const callbacks: { [k in SupportedEvents]: (context: Context, env: EnvConfigType) => Result | Promise<Result> } = {
+  "issues.closed": handleIssueClosed,
+  "issues.assigned": handleIssueAssigned,
+  "issues.unassigned": handleIssueClosed,
```
whilefoo, line 16: won't this mean that when there are 2 assignees and one gets unassigned, it will be removed from DB?

src/helpers/time.ts
```diff
@@ -0,0 +1,44 @@
...
+  });
+  const durationLabel = labels.data.find((o) => o.name.match(timeLabelRegex));
+  if (!durationLabel) {
+    return Duration.invalid("No time label was found.");
+  }
+  return parseDurationString(durationLabel.name);
+}
+
+export function parseDurationString(durationString: string) {
+  const match = durationString.match(/<(\d+)\s*(\w+)/);
```
whilefoo, line 19: I think it's better to match without `<` for extra flexibility

src/helpers/update-tasks.ts
```diff
@@ -0,0 +1,172 @@
...
+      owner,
+      repo,
+      issue_number,
+      per_page: 100,
+    },
+    (res) =>
+      res.data.filter(
+        (o) =>
+          payload.issue?.assignees?.find((assignee) => assignee?.login === o.actor.login) &&
+          DateTime.fromISO(o.created_at) >= DateTime.fromISO(issue.last_check)
```
whilefoo, line 112: personally I'd move the second part out of the function because the name of the function doesn't imply that it will also filter events since last check

src/helpers/update-tasks.ts
```diff
@@ -0,0 +1,172 @@
...
+  await octokit.rest.issues.createComment({
+    owner,
+    repo,
+    issue_number,
+    body: `@${logins}, this task has been idle for a while. Please provide an update.`,
+  });
+  return true;
+}
+
+async function removeIdleAssignees(context: Context, issue: Database["public"]["Tables"]["repositories"]["Row"]) {
```
whilefoo, line 139:
```suggestion
async function removeAllAssignees(context: Context, issue: Database["public"]["Tables"]["repositories"]["Row"]) {
```
I think this better reflects what the function does

src/adapters/index.ts
```diff
@@ -0,0 +1,67 @@
+import { SupabaseClient } from "@supabase/supabase-js";
+import { Context } from "../types/context";
+import { Database } from "../types/database";
+
+export function createAdapters(supabaseClient: SupabaseClient<Database>, context: Context) {
+  return {
+    supabase: {
+      repositories: {
```
whilefoo, line 8: `repositories` is a bit odd name because these are basically issues

whilefoo (changes requested):

src/adapters/index.ts
0x4007: The checks are being managed at the repository level. Notice the code later it checks the repository URL and then will loop down through all the issues if the conditions are fine based on the repository last check time, as I understand.

src/helpers/time.ts
```diff
@@ -0,0 +1,44 @@
...
+    case "day":
+    case "days":
+      duration = { days: parseInt(value) };
+      break;
+    case "week":
+    case "weeks":
+      duration = { weeks: parseInt(value) };
+      break;
+    default:
+      throw new Error("Unsupported duration unit.");
```
0x4007, line 40: Months might be useful for some less precise organizations.

whilefoo:
> @gentlementlegen Does the kernel support this syntax?
> 
> ```
> plugins:
>   issues.assigned:
>   issues.unassigned:
>   issues.closed:
>     - uses:
>         - plugin: ubiquibot/user-activity-watcher
>           type: github
>           with:
>             unassignUserThreshold: 5
>             sendRemindersThreshold: 1
> ```
> 
> If not, should we support it? At least it could be useful for `ubiquibot/user-activity-watcher`.

That would be difficult to implement because if there is another plugin in `issues.unassigned`, how can we know that `issues.closed` plugin wants to use all previous events.

The only solution is using this syntax:
```
plugins:
   [issues.assigned, issues.unassigned, issues.closed]:
     - uses:
         - plugin: ubiquibot/user-activity-watcher
           type: github
           with:
             unassignUserThreshold: 5
             sendRemindersThreshold: 1
```
But I think that's a problem because JS objects can't have arrays as keys but it seems that yaml library can use a Map instead of an object but that destroys type safety

gentlementlegen:
@whilefoo had that idea due to the other discussion on how to merge configurations, but wouldn't it be a solution to change the configuration to take plugins as unique keys instead? I was thinking of something like
```
plugins:
   - plugin: ubiquibot/user-activity-watcher
      type: github
      runsOn: [ 'issues.opened', 'issues.closed', 'issues.assigned' ]
      with:
        unassignUserThreshold: 5
        sendRemindersThreshold: 1
```

This way it would be very simple to reuse the plugin for different events, without needing to set it inside of the wildcard event to avoid duplicating the configuration. This would also ensure that every plugin is there only once, and would make merging probably easier.

whilefoo:
In your example you can't make a plugin chain (multiple plugins that run in order and share inputs/outputs)

If we do it like this we can still have plugins chains but there is still a problem of merging:
```yaml
plugins:
    - runsOn: [ 'issues.opened', 'issues.closed', 'issues.assigned' ]
      uses:
        - plugin: ubiquibot/user-activity-watcher
          with:
            unassignUserThreshold: 5
            sendRemindersThreshold: 1
        - plugin: ubiquibot/plugin-2
          with:
            test: true
```

src/proxy/index.ts
gentlementlegen: It is true that I didn't think about that case. I will check on `unassigned` if there is still assignees

src/helpers/get-env.ts
```diff
@@ -0,0 +1,13 @@
...
+import envConfigSchema, { envConfigValidator } from "../types/env-type";
+
+export async function getEnv() {
+  if (!envConfigValidator.test(process.env)) {
+    for (const error of envConfigValidator.errors(process.env)) {
+      console.error(error);
+    }
+    return Promise.reject(new ValidationException("The environment is invalid."));
+  }
+  return Promise.resolve(Value.Decode(envConfigSchema, process.env));
```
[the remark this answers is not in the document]
gentlementlegen: By doing so, it changes the return type to
image

I think it makes more sense to have the Promise always resolve with the env value and reject with errors no?

src/adapters/index.ts
gentlementlegen: When I started I did it repository based, but afterwards I did it issue based because it was more efficient and accurate to track by issue more than by repo. I will rename it.

supabase/migrations/20240603110253_init.sql
gentlementlegen: I removed them since they were clearly not needed.

src/worker.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+      if (!envConfigValidator.test(env)) {
+        for (const error of envConfigValidator.errors(env)) {
+          console.error(error);
+        }
+        return Promise.reject(new ValidationException("The environment is invalid."));
+      }
+      const webhookPayload = await request.json();
+      const settings = Value.Decode(userActivityWatcherSettingsSchema, Value.Default(userActivityWatcherSettingsSchema, JSON.parse(webhookPayload.settings)));
+      webhookPayload.eventPayload = JSON.parse(webhookPayload.eventPayload);
```
whilefoo, line 31:
```suggestion
```

src/worker.ts
```diff
@@ -0,0 +1,45 @@
...
+        });
+      }
+      if (!envConfigValidator.test(env)) {
+        for (const error of envConfigValidator.errors(env)) {
+          console.error(error);
+        }
+        return Promise.reject(new ValidationException("The environment is invalid."));
+      }
+      const webhookPayload = await request.json();
+      const settings = Value.Decode(userActivityWatcherSettingsSchema, Value.Default(userActivityWatcherSettingsSchema, JSON.parse(webhookPayload.settings)));
```
whilefoo, line 30:
```suggestion
      const settings = Value.Decode(userActivityWatcherSettingsSchema, Value.Default(userActivityWatcherSettingsSchema, webhookPayload.settings));
```
for workers settings and eventPayload are not even encoded anymore, only for Github actions

src/helpers/time.ts
```diff
@@ -0,0 +1,26 @@
+import { Duration } from "luxon";
+import { Context } from "../types/context";
+import ms from "ms";
+
+export async function getTimeEstimate(context: Context) {
+  const timeLabelRegex = /Time: <(\d+)/i;
```
whilefoo, line 6:
```suggestion
  const timeLabelRegex = /Time: (\d+)/i;
```

whilefoo (changes requested):

whilefoo (approved):

src/helpers/time.ts
0x4007: Although I understand your logic, the purpose of these labels are to be coarse estimates. Many project managers think in the resolution of "months" for projects, which on average, are 30 days.

0x4007:
My only feedback is to [add support for months](https://github.com/ubiquibot/user-activity-watcher/pull/1#discussion_r1643046078) and consider them 30 days.

src/helpers/time.ts
gentlementlegen: Since I changed parsing to `ms` to have a robust and battle tested library, I checked and there was a ticket opened 9 years ago on this topic here https://github.com/vercel/ms/issues/57
I will add custom logic to support months and base them on a 30 days duration.

gentlementlegen:
@0x4007 Added [custom support for months](https://github.com/ubiquibot/user-activity-watcher/pull/1/files#diff-149857a48357ac439ebf40a60e7a3497985c53b88a63c6ce40c17a7f1514767aR26) and related test.

src/helpers/time.ts
```diff
@@ -22,5 +22,14 @@ export function parseDurationString(durationString: string) {
     throw new Error("Invalid duration string format.");
   }
   const [, value] = match;
+  // Custom support for months, based on a 30 days duration
+  if (/months?/i.test(value)) {
+    const monthsMatch = value.match(/\d+/);
+    if (monthsMatch) {
+      return Duration.fromObject({ months: Number(monthsMatch[0]) }).shiftToAll();
```
0x4007, line 29: `Duration.fromObject` I am not familiar with this method, or object. I thought you said there was no support for months, so I assumed you would manually handle it as "30 days" at this line of code.

tests/main.test.ts
```diff
@@ -61,5 +61,6 @@ describe("Run tests", () => {
     expect(parseDurationString("Time: <1 Week").get("weeks")).toEqual(1);
     expect(parseDurationString("Time: <4 Hours").get("hours")).toEqual(4);
     expect(parseDurationString("Time: <8 Weeks").get("months")).toEqual(2);
+    expect(parseDurationString("Time: <3 Months").get("months")).toEqual(3);
```
0x4007, line 64: Is it worth adding tests for non plural as well? i.e. `Time: <1 Month` as well as for the other resolutions i.e. `Hour`, `Day`

src/helpers/time.ts
gentlementlegen: Correct, `ms` cannot convert months to milliseconds, however `Luxon` considers a month as 30 days (`Duration` comes from it) which is why I manually extract the duration.

tests/main.test.ts
gentlementlegen: Added singular / plural for each unit.

PR #3, opened 2024-06-13
feat: flexible timing

### LINKED ISSUE #2: Flexible timing
Depends on https://github.com/ubiquibot/user-activity-watcher/pull/1

Right now this plugin supports only thresholds in minutes:
```
- plugin: ubiquibot/user-activity-watcher
  type: github
  with:
    unassignUserThreshold: 7
    sendRemindersThreshold: 3.5
```

It would be useful to support custom string thresholds like: `1 minute`, `2 hours`, `3 days` etc... which could be converted to seconds under the hood.

Example:
```
- plugin: ubiquibot/user-activity-watcher
  type: github
  with:
    unassignUserThreshold: "1 week"
    sendRemindersThreshold: "3 days"
```

What should be done:
- take any battled tested "string to seconds" library ([example](https://www.npmjs.com/package/ms)) and add a feature of supporting string thresholds

gentlementlegen:
Resolves #2
Depends on #1 
QA run: https://github.com/Meniole/todel/issues/6


README.md
```diff
@@ -52,6 +52,6 @@ yarn test
 - plugin: ubiquibot/user-activity-watcher
   type: github
   with:
-    unassignUserThreshold: 7
-    sendRemindersThreshold: 3.5
+    unassignUserThreshold: "7 days"
```
0x4007, line 55: This value syntax is a good approach

0x4007 (approved):

src/types/plugin-inputs.ts
```diff
@@ -12,18 +13,41 @@ export interface PluginInputs<T extends WebhookEventName = SupportedEvents> {
   ref: string;
 }
 
+function thresholdType(options?: StringOptions) {
+  return T.Transform(
+    T.String({
+      // Matches a pattern like [decimal] [unit], e.g. 3.25 hours
+      pattern: /^\s*\d+(\.\d+)?\s+\S+\s*$/.source,
```
whilefoo, line 20:
```suggestion
      pattern: /^\s*\d+(\.\d+)?\s*\S+\s*$/.source,
```
`ms` also supports strings like `1h`. anyway I wouldn't even test for a pattern because `ms` handles everything

tests/main.test.ts
```diff
@@ -52,6 +54,10 @@ describe("Run tests", () => {
     await expect(getEnv()).rejects.toEqual(new ValidationException("The environment is" + " invalid."));
     process.env = oldEnv;
   });
+  it("Should parse thresholds", async () => {
```
whilefoo, line 57: maybe also a test with invalid values

whilefoo (changes requested):

0x4007:
Consider renaming the property names:

warning: "3.5 days"
disqualification: "7 days"

whilefoo (approved):

PR #8, opened 2024-07-01
feat: linked pull request activity is now taken into account

### LINKED ISSUE #6: Check PR activity
The activity watcher should also look for activity / events in the linked PR. This will prevent closing the issue if the PR is still active thus meaning the issue is still active.

gentlementlegen:
Resolves #6

### DIFF
--- a/src/handlers/collect-linked-pulls.ts
+++ b/src/handlers/collect-linked-pulls.ts
@@ -1,0 +1,78 @@
+import { parseGitHubUrl } from "../helpers/github-url";
+import { Context } from "../types/context";
+import { GitHubLinkEvent, GitHubTimelineEvent, isGitHubLinkEvent } from "../types/github-types";
+
+export type IssueParams = ReturnType<typeof parseGitHubUrl>;
+
+export async function collectLinkedPullRequests(context: Context, issue: IssueParams) {
+  const onlyPullRequests = await collectLinkedPulls(context, issue);
+  return onlyPullRequests.filter((event) => {
+    if (!event.source.issue.body) {
+      return false;
+    }
+    // Matches all keywords according to the docs:
+    // https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
+    // Works on multiple linked issues, and matches #<number> or URL patterns
+    const linkedIssueRegex =
+      /\b(?:Close(?:s|d)?|Fix(?:es|ed)?|Resolve(?:s|d)?):?\s+(?:#(\d+)|https?:\/\/(?:www\.)?github\.com\/(?:[^/\s]+\/[^/\s]+\/(?:issues|pull)\/(\d+)))\b/gi;
+    const linkedPrUrls = event.source.issue.body.match(linkedIssueRegex);
+    if (!linkedPrUrls) {
+      return false;
+    }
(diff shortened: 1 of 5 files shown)


src/handlers/collect-linked-pulls.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export type IssueParams = ReturnType<typeof parseGitHubUrl>;
+
+export async function collectLinkedPullRequests(context: Context, issue: IssueParams) {
+  const onlyPullRequests = await collectLinkedPulls(context, issue);
+  return onlyPullRequests.filter((event) => {
+    if (!event.source.issue.body) {
+      return false;
+    }
+    // Matches all keywords according to the docs:
```
0x4007, line 13: This is an interesting approach. I think it's fine. 

Should we only support issue.body (the issue specification) or should we also support comments? 

For example, within a separate discussion I could write a comment that says resolves #0 and have the comment itself (and the subsequent conversation) indicate that it's associated with the project? 

I feel like there were examples of this in the past but I can't remember any in particular right now as a reference point.

src/handlers/collect-linked-pulls.ts
0x4007, line 1: This seems like we should consider including these functions as part of our SDK instead of manually including in plugins. For now this is totally fine but if we keep importing this file then we know what we need to do.  

Seems reasonable for the bot to consider both the issue and pull for any work evaluation.

0x4007 (approved):

src/handlers/collect-linked-pulls.ts
gentlementlegen: It's a good idea, but I think it is very easy to abuse as anyone could link any comment to get credit on it. Would be worth thinking about it and see if we can make something reliable.

src/handlers/collect-linked-pulls.ts
gentlementlegen: Definitely thought about it because this was copy pasta from another plugin and it feels like a common scenario since there is no native API to achieve this (and honestly I don't understand why).

rndquu:
@gentlementlegen What is the difference between `connected` and `cross-referenced` issue events?

When PR body is updated with `Resolves #1` then the `connected` issue event is created. When issue is mentioned in the issue or PR comments then `cross-referenced` event is created. Correct?

If so then why do we need to track `cross-referenced` issue type [here](https://github.com/ubiquibot/user-activity-watcher/blob/f195b05826058d68e35f90dd236e31142346bf66/src/handlers/collect-linked-pulls.ts#L53)? Can't we simply check only for the `connected` type?

gentlementlegen:
@rndquu From my understanding, `connected` events can only be created through the Github UI (this website) when you click on the gear next to `Development` and manually add a pull-request. `cross-referenced` happens either when you write `Resolves #1` in the body of the PR, or when you write a comment saying `See also #1`, if I remember correctly.

However, according to the docs, `connected` event should be the one defining that a pull-request is supposed to close the linked issue, so theoretically you are right: https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#connected
To be double checked but I don't think that's what actually happens.

rndquu (approved):

src/handlers/collect-linked-pulls.ts
0x4007: Abuse is easy to defend against. We simply delete the offending comment before permit generation.

PR #13, opened 2024-07-28
Refactor/remove supabase

### LINKED ISSUE #10: Infinite trigger of the workflow
The root cause is not known yet, but it seems that the workflow happens to be triggered in an infinite loop (see the Action tab) when it is enabled. It might be linked with a Supabase table error, and needs to be investigated.

```
code: '42P01',
  details: null,
  hint: null,
  message: 'relation "public.issues" does not exist'
```
```
  87b49fd0-3d34-11ef-85d2-f38796c53a82 workflow_run.in_progress…2024-07-08 22:15:40
  87bb5690-3d34-11ef-966c-b04edac9dd89 workflow_job.in_progress…2024-07-08 22:15:40
  8737a7a0-3d34-11ef-972b-eedd980e4681 workflow_run.completed…2024-07-08 22:15:39
  873313c0-3d34-11ef-86cc-846d14f3abb1 check_suite.completed…2024-07-08 22:15:39
  8706ac90-3d34-11ef-98f0-06cbe010868a workflow_job.completed…2024-07-08 22:15:39
  87076fe0-3d34-11ef-876a-b6f7adf278a0 check_run.completed…2024-07-08 22:15:39
...
```

Relates to https://github.com/ubiquity/ubiquibot-kernel/issues/72

### LINKED ISSUE #12: Proposal: remove supabase
We could remove Supabase all together as everything could be done in memory using the GitHub API.

I assume it was used to cut down processing times vs parsing every open issue and set of comments every time but if this is the only reason then it should definitely be removed 

As far as I can tell supabase is being used only to "watch" issues when the same could be done with assigned and opened issues followed by a quick parsing of bot comments for the deadline.

This would resolve #10 given the error log regarding the table not existing

---

If this is run strictly as a workflow then it should be possible for us to use the run output to store a `lastCheck` value and fetch it on each new run

rfc @gentlementlegen

Keyrxng:
Resolves #12 
May Resolve #10

- removes `Env` and `adapters`
- removes supabase completely
- parses the deadline using two comments: the `/start` response string literal and the follow up comment `created_at`
- adds `ubiquibot-logger`
- new config item `watch: {optIn, optOut}`
- removed the handlers as they only served supabase functionality

Which events should be defined in `Supported Events` the same as before and just run the plugin on each or is the wildcard the real event that matters?

It would be handy for the `/start` plugin to log the assignee name in the metadata (I'll add in one of my open PRs) and the responses can be mapped to users more reliably

`time.ts` is not used in my implementation but if it's preferred/needed I can incorporate it


0x4007:
I suppose this is sort of double dipping tasks, perhaps it makes sense to consolidate into a single task #10

src/run.ts
```diff
@@ -1,41 +1,19 @@
...
-      },
-    },
-    adapters: {} as ReturnType<typeof createAdapters>,
+    logger: new Logs("info"),
   };
-  context.adapters = createAdapters(supabaseClient, context);
+
   await updateTasks(context);
-  return JSON.stringify(await proxyCallbacks[inputs.eventName](context, env));
+  return JSON.stringify({ status: "ok" }); // don't really need to return anything to the kernel
```
0x4007, line 18: Why return anything at all then? How about just a 200 status code, no body?

src/helpers/get-env.ts
```diff
@@ -1,21 +1,8 @@
...
-    for (const error of envConfigValidator.errors(process.env)) {
-      console.error(error);
-    }
-    throw new ValidationException("The environment is invalid.");
-  }
-  return Value.Decode(envConfigSchema, process.env);
-}
-
-export async function getGithubIssue(context: Context, issue: Database["public"]["Tables"]["issues"]["Row"]) {
+export async function getGithubIssue(context: Context, issue: ListIssueForRepo) {
```
0x4007, line 5:
```suggestion
export async function getGitHubIssue(context: Context, issue: ListIssueForRepo) {
```

Keyrxng:
> I suppose this is sort of double dipping tasks, perhaps it makes sense to consolidate into a single task #10

It's not guaranteed that this does resolve #10 and removing supabase was necessary anyway wasn't it? I feel they should be kept separate

src/run.ts
Keyrxng: Done, I was keeping things inline with tests

src/helpers/update-tasks.ts
```diff
@@ -1,111 +1,153 @@
...
+  const botFollowupComments = botComments.filter((o) => botFollowup.test(o?.body || ""));
+
+  if (!botAssignmentComments.length && !botFollowupComments.length) {
+    logger.info(`No assignment or followup comments found for ${issue.url}`);
+    return false;
+  }
+
+  const lastCheckComment = botFollowupComments[0]?.created_at ? botFollowupComments[0] : botAssignmentComments[0];
+
+  const lastCheck = DateTime.fromISO(new Date(lastCheckComment?.created_at).toISOString())
```
whilefoo, line 75: Is there a reason why you create a JS Date and convert to ISO string? `created_at` is already in ISO format

src/types/plugin-inputs.ts
```diff
@@ -36,6 +36,17 @@ export const userActivityWatcherSettingsSchema = T.Object({
...
+   * Define how different organizations, users or specific repositories
+   * should be watched. Use the following format:
+   * 
+   * - "ubiquibot" - all repositories in the organization
+   * - "ubiquity/ubiquibot-logger" - specific repository
+   */
+  watch: T.Object({
+    optIn: T.Array(T.String()),
+    optOut: T.Array(T.String()),
+  }),
```
whilefoo, line 49: I'm not sure if you meant to implement this new watch field entirely separate from plugins or in settings?
```
watch:
  optIn: ...

plugins:
  - uses:
    - plugin: ubiquibot/user-activity-watcher@main
      with:
        unassignUserThreshold: "5 minutes"
        sendRemindersThreshold: "1 minutes"
```
OR
```
plugins:
  - uses:
    - plugin: ubiquibot/user-activity-watcher@main
      with:
        unassignUserThreshold: "5 minutes"
        sendRemindersThreshold: "1 minutes"
        watch:
          optIn: ...
```
because the code you wrote will only work for the latter.

Also in the spec there's an example where `optIn` includes other organizations - how does it make sense for one org to turn on this plugin for another org - they might be related orgs but we don't know that?

whilefoo (changes requested):

src/worker.ts
```diff
@@ -1,11 +1,9 @@
 import { Value } from "@sinclair/typebox/value";
-import { ValidationException } from "typebox-validators";
 import { run } from "./run";
-import { EnvConfigType, envConfigValidator } from "./types/env-type";
 import { userActivityWatcherSettingsSchema } from "./types/plugin-inputs";
 
 export default {
-  async fetch(request: Request, env: EnvConfigType): Promise<Response> {
+  async fetch(request: Request): Promise<Response> {
```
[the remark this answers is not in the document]
Keyrxng: Why is this relevant to `env`, isn't the `authToken` passed in the payload and this is what's used for access?

src/types/plugin-inputs.ts
Keyrxng: Option two where it's built into the plugin settings afaik the global settings are pretty much deprecated? So it'll be defined at the org level and less likely to be customized per repo (this value at least perhaps the thresholds might be)

> - how does it make sense for one org to turn on this plugin for another org - they might be related orgs but we don't know that?

How can we validate that there is ownership between the two? 

- could we use the `installID` or would it be different for each org regardless of whether I give my bot access to all 4 of my orgs?
- even if the above is true, I see that they might still be different
- same admins/owner can't be guaranteed

We could restrict updates to only the payload organization and from this config we can tell which ones

0x4007: Installation ID corresponds to each unique organization it is added to.

src/helpers/update-tasks.ts
```diff
@@ -1,111 +1,153 @@
...
+    issue_number: issue.number,
+    per_page: 100,
+  }) as ListCommentsForIssue[];
+
+  const botComments = comments.filter((o) => o.user?.type === "Bot");
+  const dateRegex = /(?<=<td>)(\w{3}, \w{3} \d{1,2}, \d{1,2}:\d{2} (AM|PM) UTC)(?=<\/td>)/gi;
+  const assignmentRegex = /Ubiquity - Assignment - start -/gi;
+  const botAssignmentComments = sortAndReturn(botComments.filter((o) => assignmentRegex.test(o?.body || "")), "desc");
+  const botFollowup = /this task has been idle for a while. Please provide an update./gi;
+  const botFollowupComments = botComments.filter((o) => botFollowup.test(o?.body || ""));
```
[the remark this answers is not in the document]
0x4007: I had a generalized function for writing metadata that would add a "comment header" which included the function caller's name. This was intended to be used as a target for future metadata.

0x4007:
> > I suppose this is sort of double dipping tasks, perhaps it makes sense to consolidate into a single task #10
> 
> It's not guaranteed that this does resolve #10 and removing supabase was necessary anyway wasn't it? I feel they should be kept separate

I don't know enough about the problem to say definitively either way.

src/types/plugin-inputs.ts
Keyrxng: > So the opt system is intended to be used in the organization config only? Or alternatively if defined within repo config, then does it mean that only the repo it is defined in can invoke the plugin?

My understanding is we don't use global config settings anymore instead it's through each plugin's settings. I don't see why this plugin would be customized at the repo level realistically but certainly not `watch` as what would be the need?

Is the problem that if the other items are modified at the repo level then org level `watch` would be overridden?

> > same admins/owner can't be guaranteed

> Why not?

> Because we should treat every org separately

In our case yeah it can but it doesn't seem like it'd be a good blanket approach.

> So at risk of a possible security leak, we can define the installation IDs in the private config file potentially to prove ownership?

At the cost of simplicity, couldn't we encrypt it and then use their actual installID to decrypt it? This way it's stored and transmitted securely then the kernel/this plugin could verify?

src/helpers/update-tasks.ts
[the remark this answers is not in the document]
Keyrxng: That is what you are seeing here is it not?

```ts
const assignmentRegex = /Ubiquity - Assignment - start -/gi;
```

```txt
<!-- Ubiquity - Assignment - start - 86f94c4
{
  "duration": 14400,
  "priceLabel": {
    "id": 7036610424,
    "node_id": "LA_kwDOL-In888AAAABo2oneA",
    "url": "https://api.github.com/repos/ubiquibot/command-start-stop/labels/Price:%20200%20USD",
    "name": "Price: 200 USD",
    "color": "1f883d",
    "default": false,
    "description": null
  }
}
-->
```

https://github.com/ubiquibot/command-start-stop/issues/2#issuecomment-2227499243

>  It'd be better if we store data in the comment metadata but I'm not sure if we have this functionality implemented.

I forgot to last night but I will add that minor change into one of my open PRs and include the deadline, the userID etc and it can be mapped easier

src/types/plugin-inputs.ts
0x4007: Encryption aside, if the `.ubiquibot-config` repository is private then I'm assuming the data won't be leaked, right?

Keyrxng: I believe so yeah but haven't you mentioned before that repo should be public?

src/worker.ts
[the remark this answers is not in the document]
Keyrxng: I see now you mean remove `worker.ts` etc, will do

src/types/plugin-inputs.ts
0x4007: Public maybe for forking to set up a new organization but the real ones in use should be private so that we can store sensitive information if we want I suppose. 

I don't recall why I would request for them to be public.

src/helpers/update-tasks.ts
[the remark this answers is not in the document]
whilefoo: I see, this metadata is generated by another plugin, right? Ideally it should be structured like JSON or something similar.
However my point still stands for bot follow-up regex but this can be fixed in a later PR

src/types/plugin-inputs.ts
whilefoo: I think linking orgs will just complicate things. It's probably easiest to just keep this plugin at org level and specify which repos to watch or all repos. We can add linking orgs at a later stage if that's needed.

Keyrxng:
I didn't want to mess around with [`collect-linked-pulls.ts`](https://github.com/ubq-testing/user-activity-watcher/blob/refactor/remove-supabase/src/handlers/collect-linked-pulls.ts) as it seems like quite a bit of work may have went into it, it felt rude almost to just remove it and it works so no need really.

I said to whilefoo that the `search.issuesAndPullRequests` is very OP earlier and it looks like it can be used to track [linked PRs](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-for-linked-issues-and-pull-requests). I haven't tested it but I think nearly all codebases that require linkedPR tracking each has a unique way of doing it, maybe this the search endpoint is the simplest and most reliable? Thoughts on that?

0x4007:
> I didn't want to mess around with [`collect-linked-pulls.ts`](https://github.com/ubq-testing/user-activity-watcher/blob/refactor/remove-supabase/src/handlers/collect-linked-pulls.ts) as it seems like quite a bit of work may have went into it, it felt rude almost to just remove it and it works so no need really.

The link doesn't work for me but the file name sounds familiar. Im pretty sure I authored it and it's fine to delete obsolete code. 

> 
> 
> I said to whilefoo that the `search.issuesAndPullRequests` is very OP earlier and it looks like it can be used to track [linked PRs](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-for-linked-issues-and-pull-requests). I haven't tested it but I think nearly all codebases that require linkedPR tracking each has a unique way of doing it, maybe this the search endpoint is the simplest and most reliable? Thoughts on that?

If it is more reliable then we should switch to this approach.

whilefoo:
> I said to whilefoo that the `search.issuesAndPullRequests` is very OP earlier and it looks like it can be used to track [linked PRs](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-for-linked-issues-and-pull-requests).

I can see that you can filter PRs that have linked issues, but how can you specify that you only want PRs that are linked to a particular issue?

Keyrxng:
> > I said to whilefoo that the `search.issuesAndPullRequests` is very OP earlier and it looks like it can be used to track [linked PRs](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-for-linked-issues-and-pull-requests).
> 
> I can see that you can filter PRs that have linked issues, but how can you specify that you only want PRs that are linked to a particular issue?

`repo:ubiquibot/user-activity-watcher is:open linked:issue "resolves #12"` works fine. It is also case-insensitive regarding `Resolves` after manually testing via the UI. Some approaches consider other prefixes, idk how accurate it would be if we actually used the other regex captures but. 

Thoughts?

gentlementlegen:
@Keyrxng I also like using the search api because it is powerful and reliable. If it works as expected and gives the same results, it would be great to replace the current code base with it because that wound greatly reduce the code base thus lowering the amount of possible bugs.

whilefoo:
> `repo:ubiquibot/user-activity-watcher is:open linked:issue "resolves #12"` works fine. It is also case-insensitive regarding `Resolves` after manually testing via the UI. Some approaches consider other prefixes, idk how accurate it would be if we actually used the other regex captures but.
> 
> Thoughts?

I don't think this will work if the PR was linked via UI, however this already doesn't work with the current approach anyway because even though it relies on events, it still checks the comment. So I guess using search is still a better approach.

`"resolves #12"` will match any comment so we should limit to the PR description with `in:body`.
It would be also good to match other keywords. Github doesn't have any docs on boolean operators but it seems to work using OR: `is:pr is:open in:body "closes #14" OR "resolves #14" `

src/handlers/collect-linked-pulls.ts
```diff
@@ -1,78 +1,19 @@
...
-        if (url) {
-          const linkedRepo = parseIssueUrl(url);
-          isClosingPr = linkedRepo.issue_number === issue.issue_number && linkedRepo.repo === issue.repo && linkedRepo.owner === issue.owner;
-        }
-      }
-    }
-    return isGitHubLinkEvent(event) && event.source.issue.pull_request?.merged_at === null && isClosingPr;
-  });
+function additionalBooleanFilters(issueNumber: number) {
+  return `linked:${issueNumber} in:body "closes #${issueNumber}" OR "closes #${issueNumber}" OR "fixes #${issueNumber}" OR "fix #${issueNumber}" OR "resolves #${issueNumber}"`;
```
gentlementlegen, line 8:
```typescript
["closes", "fixes", "fix", "resolves"].map(o => `${o} #${issueNumber}`).join(" OR ")
```
might be less error prone. Also you do not cover all the possible keywords, see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

.gitignore
```diff
@@ -14,3 +14,5 @@ cypress/screenshots
 .dev.vars
 /tests/http/http-client.private.env.json
 .wrangler
+test-dashboard.md
+t.ts
```
gentlementlegen, line 18:
```suggestion

```

src/helpers/update-tasks.ts
```diff
@@ -1,120 +1,182 @@
...
+    owner: repo.owner.login,
+    repo: repo.name,
+    issue_number: issue.number,
+    per_page: 100,
+  }) as ListCommentsForIssue[];
+
+  const botComments = comments.filter((o) => o.user?.type === "Bot");
+  const taskDeadlineJsonRegex = /"taskDeadline": "([^"]*)"/g;
+  const taskAssigneesJsonRegex = /"taskAssignees": \[([^\]]*)\]/g;
+  const assignmentRegex = /Ubiquity - Assignment - start -/gi;
```
gentlementlegen, line 62: This seems very risky if we ever change any message. Is there no better way?

src/helpers/update-tasks.ts
```diff
@@ -1,120 +1,182 @@
...
+  } else if (metadata.taskAssignees.length && issue.assignees?.length && metadata.taskAssignees.some((a) => !issue.assignees?.map((o) => o.id).includes(a))) {
+    logger.error(`Assignees mismatch found for ${issue.url}`);
+    return false;
+  }
+
+  if (!metadata.taskDeadline) {
+    logger.info(`No deadline found for ${issue.url}`);
+    return false;
+  }
+  const deadline = DateTime.fromFormat(metadata?.taskDeadline, "EEE, LLL d, h:mm a 'UTC'")
```
gentlementlegen, line 100: Why not using ISO anymore?

src/helpers/update-tasks.ts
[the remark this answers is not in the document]
gentlementlegen: I just saw this thread, I had the same concern myself
https://github.com/ubiquibot/user-activity-watcher/pull/13#discussion_r1721376180

Even by using metadata checking for a string seems unreliable and easily breakable. What is the purpose of checking for this bot comment?

src/helpers/update-tasks.ts
0x4007: The intent behind the metadata was to be able to target at least the developer set "metadata class name" which in this case seems to be "Assignment".

gentlementlegen: I have to test it to see the logic but is this used to updated a previous comment? Because if so, I don't think you get notifications for this.

src/run.ts
```diff
@@ -1,41 +1,19 @@
...
-      },
-      error(message: unknown, ...optionalParams: unknown[]) {
-        console.error(message, ...optionalParams);
-      },
-      fatal(message: unknown, ...optionalParams: unknown[]) {
-        console.error(message, ...optionalParams);
-      },
-    },
-    adapters: {} as ReturnType<typeof createAdapters>,
+    logger: new Logs("info"),
```
gentlementlegen, line 14: Maybe debug would be better, or make it configurable. It is useful to have all the logs for debugging, even more on plugins like this one.

gentlementlegen:
@Keyrxng We used to have a different timer for reminders and disqualifications if the user was either a collaborator or on contributor, is this changed?

src/helpers/update-tasks.ts
0x4007: We can certainly do a test. Just message me on telegram when to check a test notification.

gentlementlegen: Whenever I try to run it I get `Missing metadata from ...`
https://github.com/Meniole/user-activity-watcher/actions/runs/10504703437/job/29100696449#step:5:61

src/handlers/collect-linked-pulls.ts
Keyrxng: https://github.com/ubiquibot/user-activity-watcher/pull/15

src/helpers/update-tasks.ts
Keyrxng: It's not to update a previous comment. This just reads the bot comment that's posted when someone uses `/start` so we can extract the metadata from it

Keyrxng: Also the updated comment has not been merged yet so it's relying on V1 

https://github.com/ubiquibot/command-start-stop/pull/19/files#diff-47224fac5bab974df0f6fb9704d6fbb0f06375b3e98285797826104537bda3cc

in #15 I have split the functions up and you should not get that error anymore

- https://github.com/ubq-testing/user-activity-watcher/tree/chore/testing/src/helpers

gentlementlegen (approved):
I am fine with this, I guess it will be way more performant once we get #15 in.

whilefoo (approved):

PR #15, opened 2024-08-05
Chore/testing

### LINKED ISSUE #14: Improve tests
They would benefit from being brought inline with other plugin test suites covering more of the logic and paths

Time - 2 hours

Keyrxng:
Resolves #14
Requires #13


.gitignore
```diff
@@ -14,3 +14,5 @@ cypress/screenshots
 .dev.vars
 /tests/http/http-client.private.env.json
 .wrangler
+test-dashboard.md
+t.ts
```
gentlementlegen, line 18: What is this file?

Keyrxng: My mistake I'll remove that, I use `npx tsx t.ts` to run test scripts

src/helpers/update-tasks.ts
```diff
@@ -1,120 +1,182 @@
...
-      await supabase.issues.delete(issue.url);
-    }
+  const repos = await getWatchedRepos(context);
+
+  if (!repos?.length) {
+    logger.info("No watched repos have been found, no work to do.");
+    return false;
   }
+
+  for (const repo of repos) {
```
gentlementlegen, line 18: Small optimization but this could all be done in parallel through a `Promise.all` isn't it?

Keyrxng: I wouldn't as it drastically increases the chance we get rate limited doesn't it?

So we either have 1k or 15k p/h if we use secrets and if we use the `inputs.authToken` we have 5k - 15k, I think this is probably okay but it's the [secondary](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits) that we'll likely hit

> Make too many concurrent requests. No more than 100 concurrent requests are allowed. This limit is shared across the REST API and GraphQL API.
Make too many requests to a single endpoint per minute. No more than 900 points per minute are allowed for REST API endpoints

gentlementlegen: Yeah makes sense, let's keep it that way then.

tests/main.test.ts
```diff
@@ -1,86 +1,224 @@
...
-    const result = await run(program, Value.Decode(envConfigSchema, process.env));
-    expect(JSON.parse(result)).toEqual({ status: "ok" });
+    const context = createContext(1, 1);
+    const result = await runPlugin(context);
+    expect(result).toBe(true);
+  });
+
+  it("Should process update for all repos except optOut", async () => {
+    const context = createContext(1, 1);
+    const infoSpy = jest.spyOn(context.logger, "info");
```
whilefoo, line 56: I'd try to avoid making tests relying on the logger since that can change a lot. It'd be better to spy on functions or check that http mocks have been called

Keyrxng: I understand where you are coming from with this but the reason why I use the logger in some tests and then only state in others is because I figured that technically the logs are an important feature of our plugins and should probably be covered 

I've written all my test suites this way until now and they are pretty reliant on the logger. Unsure whether you meant this as advice for future unit tests or if I should consider refactoring all of them

---

to confirm, the better approach would be to stub functions which result in no state change to assert they were called and then all others can be confirmed via mock db entries, ignoring logs and confirming input/output essentially?

package.json
```diff
@@ -87,5 +85,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
-}
+  },
+  "packageManager": "yarn@1.22.22"
```
0x4007, line 89:
```suggestion

```

src/index.ts
```diff
@@ -1,10 +1,8 @@
...
-import { getEnv } from "./helpers/get-env";
 import program from "./parser/payload";
 import { run } from "./run";
 
-getEnv()
-  .then((env) => run(program, env))
+run(program)
   .then((result) => {
     core?.setOutput("result", result);
   })
```
0x4007, line 8:
```suggestion
run(program).then((result) => core?.setOutput("result", result))
```

0x4007:
How is this pull coming along?

package.json
Keyrxng: I think we should keep this and standardize it. There is no downside to including but omitting creates room for error. Why would you prefer it removed?

Keyrxng:
> How is this pull coming along?

I'm not sure whether [this](https://github.com/ubiquibot/user-activity-watcher/pull/15#discussion_r1712670770) is a blocker or not but I feel good with how things are code wise. 

I will produce manual QA for PRs which don't have any later today, this being one of them

Keyrxng:
The search param seems to be doing the damage.

- QA [run](https://github.com/ubq-testing/user-activity-watcher/actions/runs/10426649399/job/28879981991)
- [Issue](https://github.com/ubq-testing/task_xp_guard/issues/1) I was removed from for being over the limit

I got errors using more that 5 bool operators in the query so we need to keep that in mind using that endpoint. I included parsing of V1 comments which we can remove once they are fully deprecated

package.json
0x4007: Then its fine I guess but seems unnecessary because my yarn never did that

0x4007:
When do you think we can start using this plugin?

Keyrxng:
I will tidy the code up and find any final optimizations or changes I can make this morning. I'll try to create more manual QA but I have linked both me being kicked and [being reminded](https://github.com/ubq-testing/bot-ai/issues/50#issuecomment-2295861154)

Should I refactor all tests in order to get this merged, address in another PR or it's not an issue?

Keyrxng:
@gentlementlegen 

> This seems very risky if we ever change any message. Is there no better way?

What would you suggest? I am relying on the metadata created by another plugin for V2 and in terms of V1 the metadata also logs a `duration` var which is the time label so there's no need for us to calculate it from the labels.

Otherwise we just don't use the metadata from the other plugin and we just build everything from the issue details? Would that be preferred?

gentlementlegen:
@Keyrxng I think this would be way more reliable than manually parsing of course. However, I don't think Octokit ships GraphQl types, so we should use tools like [codegen](https://the-guild.dev/graphql/codegen) to generate types and ensure auto completion as well. I would be in favor to use GrqphQl, but it might require a bit more setup.

Keyrxng:
> @Keyrxng I think this would be way more reliable than manually parsing of course. However, I don't think Octokit ships GraphQl types, so we should use tools like [codegen](https://the-guild.dev/graphql/codegen) to generate types and ensure auto completion as well. I would be in favor to use GrqphQl, but it might require a bit more setup.

I'll refactor and use the GraphQL API then no worries

gentlementlegen:
Let me know if you encounter major issues. We have another repository where we also fetch all the related pull-requests, so if that is working nice we can also update the other repo afterwards.

Keyrxng:
> Let me know if you encounter major issues. We have another repository where we also fetch all the related pull-requests, so if that is working nice we can also update the other repo afterwards.

Will do thanks

Keyrxng:
@gentlementlegen the difficult part was figuring out you could mock entire modules just by creating a duplicate in `__mocks__`. I've had this problem for a day or so with `timeline-event-rewards` as Jest has troubles picking up the `@octokit/graphql-schema` package but eslint, ts etc work fine. I was close to pinging you but got there in the end, I'm glad to have found a solution here so I can start writing tests for it

0x4007:
> I will tidy the code up and find any final optimizations or changes I can make this morning. I'll try to create more manual QA but I have linked both me being kicked and [being reminded](https://github.com/ubq-testing/bot-ai/issues/50#issuecomment-2295861154)
> 

Great to see this

> 
> Should I refactor all tests in order to get this merged, address in another PR or it's not an issue?

Sure it can be handled in a separate task if it's gonna take awhile. 

Looking forward to seeing this plugin in action. It will free up many of our tasks.

tests/__mocks__/@octokit/graphql-schema.js
gentlementlegen, line 1: I guess this doesn't work as a TypeScript file? Couldn't find anything in the docs.

src/helpers/task-deadline.ts
```diff
@@ -0,0 +1,49 @@
...
+    const { logger, config } = context;
+
+    const assigneeIds = issue.assignees?.map((o) => o.id) || [];
+
+    if (assigneeIds.length && metadata.taskAssignees?.some((a) => !assigneeIds.includes(a))) {
+        logger.info(`Assignees mismatch found for ${issue.html_url}`, {
+            metadata,
+            assigneeIds,
+        });
+        // return false;
```
gentlementlegen, line 24:
```suggestion
```

src/handlers/collect-linked-pulls.ts
```diff
@@ -1,78 +1,65 @@
...
-      // Only add to connections if there is no corresponding disconnected event
-      if (!disconnections.has(issueNumber)) {
-        connections.set(issueNumber, issueEvent);
-      }
-    } else if (issueEvent.event === "disconnected") {
-      disconnections.set(issueNumber, issueEvent);
-      // If a disconnected event is found, remove the corresponding connected event
-      if (connections.has(issueNumber)) {
-        connections.delete(issueNumber);
+const query = `query collectLinkedPullRequests($owner: String!, $repo: String!, $issue_number: Int!) {
```
gentlementlegen, line 18:
```suggestion
const query = /* GraphQl */ `query collectLinkedPullRequests($owner: String!, $repo: String!, $issue_number: Int!) {
```

gentlementlegen:
It would be nice to have Graphql config, similar to the one mentioned [here](https://github.com/ubiquibot/task-xp-guard/pull/1#issuecomment-2297660301).

tests/__mocks__/@octokit/graphql-schema.js
Keyrxng: I assumed it wouldn't but it works with `.ts`

tests/__mocks__/@octokit/graphql-schema.ts
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    validate: () => [],
```
gentlementlegen, line 2: Is this just to mock the module?
Doesn't seem properly indented.

Keyrxng: Yeah Jest fails to import it. My format-on-save was forcing the indent issue for some reason

tests/main.test.ts
```diff
@@ -1,62 +1,226 @@
...
 
   it("Should parse thresholds", async () => {
     const settings = Value.Decode(userActivityWatcherSettingsSchema, Value.Default(userActivityWatcherSettingsSchema, cfg));
-    expect(settings).toEqual({ warning: 302400000, disqualification: 604800000, watch: { optOut: ["private-repo"] } });
+    expect(settings).toEqual({ warning: 302400000, disqualification: 604800000, watch: { optOut: [STRINGS.PRIVATE_REPO_NAME] } });
     expect(() =>
       Value.Decode(
         userActivityWatcherSettingsSchema,
         Value.Default(userActivityWatcherSettingsSchema, {
           warning: "12 foobars",
```
gentlementlegen, line 41: Maybe this default value should be changed to something meaningful

Keyrxng: Which default value do you mean sorry? `optOut`? This was one of the original tests the plugin had

gentlementlegen: I just understood that it is testing weird values on purpose, didn't have the full context of the test, apologies.

gentlementlegen (approved):
Testing on my org and seems to work properly: https://github.com/Meniole/user-activity-watcher/actions/runs/10546334364/job/29217677141#step:5:31

src/helpers/task-metadata.ts
```diff
@@ -0,0 +1,82 @@
...
+    const taskDeadlineJsonRegex = /"taskDeadline": "([^"]*)"/g;
+    const taskAssigneesJsonRegex = /"taskAssignees": \[([^\]]*)\]/g;
+    const assignmentRegex = /Ubiquity - Assignment - start -/gi;
+    const botAssignmentComments = botComments.filter(
+        (o) => assignmentRegex.test(o?.body || "")
+    ).sort((a, b) =>
+        DateTime.fromISO(a.created_at).toMillis() - DateTime.fromISO(b.created_at).toMillis()
+    )
+
+    const botFollowup = /this task has been idle for a while. Please provide an update./gi;
```
0x4007, line 30: Match with metadata header instead

Keyrxng: I'll need to take the function from `command-start-stop` and create a new comment as it's posted by this plugin

src/helpers/task-metadata.ts
```diff
@@ -0,0 +1,82 @@
+import { DateTime } from "luxon";
+import { Context } from "../types/context";
+import { ListCommentsForIssue, ListForOrg, ListIssueForRepo } from "../types/github-types";
+import ms from "ms";
+
+export async function handleCommentsAndMetadata(
```
whilefoo, line 6:
```suggestion
export async function getTaskMetadata(
```

src/helpers/task-deadline.ts
```diff
@@ -0,0 +1,48 @@
+import { DateTime } from "luxon";
+import { Context } from "../types/context";
+import { ListIssueForRepo } from "../types/github-types";
+import { getAssigneesActivityForIssue } from "./get-assignee-activity";
+
+export async function handleDeadline(
```
whilefoo, line 6:
```suggestion
export async function getDeadlineWithThreshold(
```

whilefoo (changes requested):

src/helpers/get-assignee-activity.ts
```diff
@@ -0,0 +1,38 @@
...
+        });
+        issueEvents.push(...events);
+    }
+
+    return issueEvents
+        .reduce((acc, event) => {
+            if (event.actor && event.actor.id) {
+                if (assigneeIds.includes(event.actor.id)) acc.push(event);
+            }
+            return acc;
```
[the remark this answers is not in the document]
Keyrxng: Yeah I guess it now that you point it out lmao

whilefoo (approved):

tests/main.test.ts
```diff
@@ -119,7 +120,9 @@ describe("User start/stop", () => {
     expect(updatedIssue?.assignees).toEqual([{ login: STRINGS.USER, id: 2 }]);
 
     const comments = db.issueComments.getAll();
-    expect(comments[comments.length - 1].body).toEqual(JSON.stringify({ body: `@${STRINGS.USER}, this task has been idle for a while. Please provide an update.` }));
+    const latestComment = comments[comments.length - 1];
+    const partialComment = "@user2, this task has been idle for a while. Please provide an update.\\n\\n\\n<!-- Ubiquity - Followup -"
```
0x4007, line 124: Perhaps it makes sense to use the real metadata writing method instead of manually trying to write the metadata.

gentlementlegen:
Interesting that the bot didn't care about your comments @0x4007 for activity. Also, the `skipBotEvents` didn't seem to work, no rewards have been generated.

0x4007:
I don't understand why reliability is decreasing

gentlementlegen:
It is using lots of new features we never tried before. Nothing to do with `conversation-rewards` capabilities.

src/helpers/task-metadata.ts
```diff
@@ -27,7 +27,7 @@ export async function handleCommentsAndMetadata(
         DateTime.fromISO(a.created_at).toMillis() - DateTime.fromISO(b.created_at).toMillis()
     )
 
-    const botFollowup = /this task has been idle for a while. Please provide an update./gi;
+    const botFollowup = /<!-- Ubiquity - Followup - remindAssignees/gi;
```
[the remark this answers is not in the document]
Keyrxng: Can you link to it I don't think I've seen it before

PR #16, opened 2024-08-19
feat: manifest file

gentlementlegen:


manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "User activity watcher",
+  "description": "Watches user activity on issues, sends reminders on deadlines, and unassign inactive users.",
+  "ubiquity:listeners": [ "push", "pull_request.opened", "pull_request.reopened" ]
```
0x4007, line 4: Should be on comments because those are the most frequent events.

gentlementlegen: I can make it just for comments then, will change.

0x4007:
I'm sorry but I just remembered we also watched for commits! So if you can add `push` again.

0x4007 (commented):
chore: add `push`

0x4007 (approved):

Keyrxng:
We still need to define `runsOn: [webhook]` for each plugin don't we? And this can differ from the manifest, isn't that right?

But we also need to ensure the `SupportedEventsU` contains everything that's in the manifest at least yeah?

What happens when `runsOn` is different from the manifest and is not in `SupportedEventsU` is that the plugin won't run if we are constructing context with a generic using `SupportedEventsU`. So should we be using the full `WebhookEventNames` union in plugins then to be able to construct all payload types? And if we do that, don't we need to typeguard them all to ensure we pull the right data from each?

Because in this code base we have

```ts
export type SupportedEvents = "issues.closed" | "issues.assigned" | "issues.unassigned";
```

but the manifest now includes:

```yml
"ubiquity:listeners": [ "pull_request_review_comment.created", "issue_comment.created", "push" ]
```

gentlementlegen:
@Keyrxng `runsOn` doesn't need to be defined. When not defined, it uses the `manifest`. When defined, it overrides the manifest. 

Supported events are made for the code to handle these events specifically. But actually the plugin can run on any events. The thing is checking the `supportedEvents` is important because the payload can be different according to the event type. But for example on the automated merging plugin, any event would be valid as long as they can refer to an org or and owner.

PR #20, opened 2024-09-12
event assignment default

### LINKED ISSUE #19: Wrong "start time" target
It appears as though the software checks for the metadata from the /start comment. 

This is incorrect because there are various ways to assign people without that command.

Check the timeline for assign events and target this timestamp. 

The correct logic can be lifted from bot v1. 

Example of it not working as expected: https://github.com/ubiquibot/conversation-rewards/issues/92

Keyrxng:
Resolves #19

- fallback to assignment event timestamp if the bot has not intervened since the contributor started the task via any method.
- updated tests


0x4007 (commented):
Where's QA link? Code looks fine.

Keyrxng:
> Where's QA link? Code looks fine.

My bad

QA:

https://github.com/ubq-testing/user-activity-watcher/actions/runs/10834775241/job/30064760402
https://github.com/ubq-testing/sandbox/issues/2

image

0x4007:
Is there a simple way you can prove that this works in some mock example with shorter timeouts?

Keyrxng:
> Is there a simple way you can prove that this works in some mock example with shorter timeouts? Like link a GitHub issue

https://github.com/ubq-testing/command-start-stop/issues/19 < this was created on the 4th, so last week. My config settings are the default `3.5` and `7` days respectively.



image

gentlementlegen:
I think what we are trying to solve is users that self assign through the UI. There is no start message from the bot (there is, but it is different). In this scenario I noticed that the reminders and unassigns were not working.
Example: https://github.com/ubiquibot/user-activity-watcher/actions/runs/10835091739/job/30065821594#step:5:53

Keyrxng:
image
image

Keyrxng:
Here is a run I just done with a `5m` warning and `10m` disqualification 

https://github.com/ubq-testing/user-activity-watcher/actions/runs/10835251567/job/30066345978
image

src/helpers/task-metadata.ts
```diff
@@ -18,65 +18,106 @@ export async function getTaskMetadata(
...
+        );
 
+    // `lastCheck` represents the last time the bot intervened in the issue, separate from the activity tracking of a user.
     const lastCheckComment = botFollowupComments[0]?.created_at ? botFollowupComments[0] : botAssignmentComments[0];
-    const lastCheck = DateTime.fromISO(lastCheckComment.created_at);
+    let lastCheck = lastCheckComment?.created_at ? DateTime.fromISO(lastCheckComment.created_at) : null;
+
+    // if we don't have a lastCheck yet, use the assignment event
+    if (!lastCheck) {
+        console.log("No last check found, using assignment event");
```
gentlementlegen, line 42: Should use the logger, and indent seems wrong.

gentlementlegen (changes requested):
The code seems to be indented with 4 spaces instead of the usual 2 spaces.

0x4007:
Fix CI and lets merge.

Keyrxng:
fixing ci rn, new log means update tests. Tedious but essential

> The code seems to be indented with 4 spaces instead of the usual 2 spaces.

Yeah idk why that happens in certain repos, I always have `format-on-save` enabled and it works right most of the time. It's not like there is a config file in this repo that's affecting it either, weird af.

PR #25, opened 2024-09-25
feat: schema validation

gentlementlegen:
Resolves https://github.com/ubiquity/ubiquity-os-kernel/issues/80


src/helpers/validator.ts
```diff
@@ -0,0 +1,50 @@
...
+    return { decodedEnv, decodedSettings };
+  } catch (e) {
+    if (e instanceof TransformDecodeCheckError || e instanceof TransformDecodeError) {
+      throw { errors: [e.error] };
+    }
+    throw e;
+  }
+}
+
+export async function returnDataToKernel(repoToken: string, stateId: string, output: object, eventType = "return_data_to_ubiquibot_kernel") {
```
0x4007, line 39:
```suggestion
export async function returnDataToKernel(repoToken: string, stateId: string, output: object, eventType = "return-data-to-ubiquity-os-kernel") {
```

We should deprecate "ubiquibot" and also underscores seem unusual in a javascript context

.github/workflows/update-configuration.yml
```diff
@@ -0,0 +1,59 @@
...
+
+              const updatedManifest = JSON.stringify(manifest, null, 2);
+              console.log('Updated manifest:', updatedManifest);
+              await fs.writeFile(manifestPath, updatedManifest);
+            })();
+
+      - name: Commit and Push generated types
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
```
0x4007, line 50: Auth as our bot using the github action in the previous step as we do in many other actions to get the auth token based on the `APP_ID` and `APP_PRIVATE_KEY`

0x4007 (approved):

PR #28, opened 2024-10-08
October Sync

0x4007:


package.json
```diff
@@ -2,14 +2,14 @@
   "name": "user-activity-watcher",
   "version": "1.0.0",
   "description": "Watches user activity on issues, sends reminders on deadlines, and unassign inactive users.",
-  "main": "src/worker.ts",
+  "main": "src/index.ts",
   "author": "Ubiquity DAO",
   "license": "MIT",
+  "type": "module",
```
0x4007, line 8: What's this for?

package.json
```diff
@@ -87,5 +88,6 @@
     "extends": [
       "@commitlint/config-conventional"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
0x4007, line 92: This shouldn't be here.

src/types/plugin-inputs.ts
```diff
@@ -31,11 +32,18 @@ function thresholdType(options?: StringOptions) {
...
    * Delay to send reminders. 0 means disabled. Any other value is counted in days, e.g. 1,5 days
    */
   warning: thresholdType({ default: "3.5 days" }),
+  /**
+   * By default all repositories are watched. Use this option to opt-out from watching specific repositories
+   * within your organization. The value is an array of repository names.
+   */
+  watch: T.Object({
+    optOut: T.Array(T.String()),
+  }),
```
0x4007, line 46: Why is this nested? It should only be optOut on the top level.

tests/__mocks__/strings.ts
```diff
@@ -0,0 +1,85 @@
+export const STRINGS = {
+  UBIQUITY: "ubiquity",
+  UBIQUIBOT: "ubiquibot",
```
0x4007, line 3: All UbiquiBot references should be removed.

package.json
gentlementlegen: This is very likely to be compatible with ESM libraries, like `octokit/graphql`.

gentlementlegen:
@0x4007 Should a task be created for the fixes?

0x4007:
Sure that's fine

PR #33, opened 2024-10-14
fix: inputs and build

### LINKED ISSUE #32: Fix default for eventWhitelist and Decode
When left blank, the configuration complains about missing `eventWhitelist`. When giving values, it seems to decode them improperly, cf run : https://github.com/ubiquity-os-marketplace/daemon-disqualifier/actions/runs/11316394599/job/31468631289#step:5:93

gentlementlegen:
Resolves #32 
QA: https://github.com/Meniole/user-activity-watcher/actions/runs/11320647090


.github/workflows/update-configuration.yml
```diff
@@ -1,59 +1,17 @@
...
-            git commit -m "chore: updated generated configuration" || echo "Lint-staged check failed"
-            git push origin HEAD:${{ github.ref_name }}
-          else
-            echo "No changes to commit"
-          fi
+      - uses: ubiquity-os/action-deploy-plugin@main
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          app-id: ${{ secrets.APP_ID }}
+          app-private-key: ${{ secrets.APP_PRIVATE_KEY }}
```
0x4007, line 17: Kebab case syntax? Is that right?

action.yml
```diff
@@ -1,16 +1,22 @@
...
+  eventName:
+    description: "Event Name"
+  eventPayload:
+    description: "Event Payload"
+  settings:
+    description: "Settings"
+  authToken:
+    description: "Auth Token"
+  ref:
+    description: "Ref"
```
0x4007, line 15: This could be more specific

.github/workflows/update-configuration.yml
gentlementlegen: It's usually prettier to use within Actions, but can be changed to camelCase if you want.

0x4007: Yeah I would consider these secret names to be standardized to `APP_ID` and `APP_PRIVATE_KEY` at this point

dist/package.json
0x4007, line 1: This seems unusual

0x4007 (approved):
Code looks fine.

src/run.ts
```diff
@@ -12,8 +12,9 @@ export async function run(inputs: PluginInputs) {
     payload: inputs.eventPayload,
     config: inputs.settings,
     octokit,
-    logger: new Logs("verbose"),
+    logger: new Logs("debug"),
```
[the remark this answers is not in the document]
gentlementlegen: Yes although the SDK integrates the logger as well so might now be worth changing there right now, I will try to switch this to the SDK later as well.

dist/package.json
gentlementlegen: When using `type: module` ncc includes this as an ESM reminder from what I've seen.
https://github.com/vercel/ncc?tab=readme-ov-file#usage-1

PR #36, opened 2024-10-21
fix: improvements on follow-up

### LINKED ISSUE #23: Follow up enhancements
1. Follow up should be on pull request so that they don't reply under issue and get paid for it. 
2. If there isn't a pull request open in time for the warning, automatically disqualify them (they should open a draft pull as soon as they self assign)
   - Must be configurable since some tasks are not coding related. 
   - If pull isn't required then follow up under issue as we do now. 
3. "activity which should be considered is only the one from the assignees themselves. I believe currently any activity from anyone is considered in the events."

gentlementlegen:
Resolves #23
QA: https://github.com/Meniole/user-activity-watcher/issues/2


.github/workflows/compute.yml
```diff
@@ -16,7 +16,7 @@ on:
       ref:
         description: "Ref"
       signature:
-        description: "The kernel signature"
+        description: "Kernel payload signature"
```
0x4007, line 19:
```suggestion
        description: "Used for authenticating requests from the kernel."
```

src/handlers/watch-user-activity.ts
```diff
@@ -1,27 +1,28 @@
...
-  }
+  await Promise.all(
+    repos.map(async (repo) => {
+      logger.debug(`> Watching user activity for repo: ${repo.name} (${repo.html_url})`);
+      await updateReminders(context, repo);
+    })
+  );
 
-  return true;
+  return { message: "OK" };
```
0x4007, line 22: How about a statusCode: 200 instead

src/helpers/task-update.ts
```diff
@@ -1,45 +1,90 @@
...
 
-  if (now >= deadlineWithThreshold) {
-    // if the issue is past due, we should unassign the user
-    await unassignUserFromIssue(context, issue);
-  } else if (now >= reminderWithThreshold) {
-    // if the issue is within the reminder threshold, we should remind the assignees
-    await remindAssigneesForIssue(context, issue);
+  const activityEvent = (await getAssigneesActivityForIssue(context, issue, handledMetadata.taskAssignees))
+    .filter((o) => eventWhitelist.includes(o.event as TimelineEvent))
+    .shift();
```
0x4007, line 46: This looks like this can shift an `undefined` value, I don't think this is type safe. Check if its undefined and throw a specific error for that if you must.

src/helpers/task-update.ts
```diff
@@ -1,45 +1,90 @@
...
+
+  const assignedDate = DateTime.fromISO(assignedEvent.created_at);
+  const activityDate = activityEvent?.created_at ? DateTime.fromISO(activityEvent.created_at) : undefined;
```
0x4007, line 49: I think it should never be undefined. You should throw for that.

src/helpers/task-update.ts
gentlementlegen: `undefined` is checked, the shift is on purpose, to check if there was any activity, and if not define the last activity as when the user got assigned.

src/helpers/task-update.ts
gentlementlegen: `activityDate` can be undefined if the user just assigned itself and never did anything afterwards (which is a common scenario in our repos). When this happens the last activity points to when the user was assigned (coincides with the user sending a `/start` command).

src/handlers/watch-user-activity.ts
gentlementlegen: Sure, will change it. Thought it was not useful because the network call already sends a `200`, this is the body.

gentlementlegen:
@whilefoo @Keyrxng if you want to have a look, I can't assign you as reviewers.

Keyrxng:
> @whilefoo @Keyrxng if you want to have a look, I can't assign you as reviewers.

Just about to clock off bud but I'm happy to after I wake.

0x4007 (approved):

gentlementlegen:
I re-enabled the workflow as well.

gentlementlegen:
I re-enabled the workflow as well.

PR #54, opened 2024-11-27
feat: deadline-message

### LINKED ISSUE #44: Disqualified early
This is the [issue](https://github.com/ubiquity-os/ubiquity-os-kernel/issues/166) where I got disqualified 3 days earlier than the deadline

When I started the task with `/start` on 28th October, the bot said the deadline is Sat, Nov 2, 2:36 PM UTC, however on 30th October, the bot said the deadline has passed and unassigned me

Found another [issue](https://github.com/ubiquity-os/permit-generation/issues/71#issuecomment-2448881918) with the same problem

gentlementlegen:
Resolves #44
QA: https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2504353561


gentlementlegen:
@0x4007 Ready when you are

dist/index.js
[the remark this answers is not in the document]
gentlementlegen: I will, but linguist does not hide the file, it just collapses it by default which is already the case because the file is large

0x4007 (changes requested):
The message needs work to accommodate both scenarios. 

Scenario 1: no pull. 
2: normal follow ups. 

We need to express the schedule to the assignee for both scenarios. 

https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2513618073

gentlementlegen:
@0x4007 When no pull request is required the sentence
`Be careful! If no pull request has been opened before the first reminder, you will automatically get disqualified.`
won't be displaying, isn't that ok?

0x4007:
> @0x4007 When no pull request is required the sentence
> 
> `Be careful! If no pull request has been opened before the first reminder, you will automatically get disqualified.`
> 
> won't be displaying, isn't that ok?
> 
> 
> 
> https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/54/files#diff-125c8ed36b16b66c130a1e01fe9c41d5a675c797e734e75d775c214664ff6eb6R31

This approach works just adapted to my new prompts as described in the QA

gentlementlegen:
@0x4007 new QA: https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2513677362

gentlementlegen:
@0x4007 updated QA to reflect your requested changes
https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2519338307

gentlementlegen:
@0x4007 new QA https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2519463850

0x4007:
You need to highlight it as monospace text copy my template. Lastly it would be nice to place an "and" between time delimiters like hours and minutes.

gentlementlegen:
@0x4007 I saw no difference on the macbook, are you using your phone? For the "and" I am using a package https://www.npmjs.com/package/pretty-ms which doesn't offer that option, so I guess I should manually parse time to achieve so.

gentlementlegen:
@0x4007 Updated QA: https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2522024384

.gitattributes
```diff
@@ -0,0 +1 @@
+dist/* linguist-generated
```
0x4007, line 1:
```suggestion
dist/** linguist-generated
```

0x4007: Bun lock file also

dist/schema.graphql
0x4007, line 1: Is this correct

manifest.json
```diff
@@ -1,15 +1,14 @@
 {
-  "name": "Daemon Disqualifier",
-  "description": "Watches user activity on issues, sends reminders on deadlines, and unassign inactive users.",
-  "ubiquity:listeners": ["pull_request_review_comment.created", "issue_comment.created", "push"],
+  "name": "User activity watcher",
```
0x4007, line 2: I thought it's supposed to match the repository name for the plugin installer or something.

gentlementlegen: Good catch, my mistake when merging the JSON from the main branch.

dist/schema.graphql
gentlementlegen: It will eventually be re-generated, it is included by octokit graphql.

.gitattributes
0x4007: You need to include all the generated files here.

gentlementlegen: What generated files are there other than the lockfile and dist folder?

0x4007: Off hand I think thats it but the lockfile is not marked as generated.

gentlementlegen: https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/54/files#diff-618cd5b83d62060ba3d027e314a21ceaf75d36067ff820db126642944145393eR2

it is in the linguist file

gentlementlegen:
@whilefoo I think you believe you were disqualified earlier than expected because the posted message about the deadlines was incorrect.

The deadline takes the priority into account, which means that 7 days will become 2.4 days on a priority 3 task, for example. Yet, the message told that you would get a reminder after 7 days. This is properly displayed here: https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2519463850

Let me just fix the duration calculation because it doesn't seem correct in the test.

src/worker.ts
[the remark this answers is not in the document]
gentlementlegen: This is to allow local run of the plugin. It is quite helpful for test / debug. I should document that somewhere, but I think the easiest way to test plugins is to run them as servers locally with `bun` (or wrangler), then use `ngrok` to redirect the network. This way I don't need to care about the kernel and can test directly within GitHub, without needed to deploy it.

src/helpers/remind-and-remove.ts
```diff
@@ -96,9 +96,12 @@ async function removeAllAssignees(context: ContextPlugin, issue: ListIssueForRep
     return false;
   }
   const logins = issue.assignees.map((o) => o?.login).filter((o) => !!o) as string[];
-  const logMessage = logger.info(`Passed the deadline and no activity is detected, removing assignees: ${logins.map((o) => `@${o}`).join(", ")}.`, {
-    issue: issue.html_url,
-  });
+  const logMessage = logger.info(
+    `Passed the disqualification threshold and no activity is detected, removing assignees: ${logins.map((o) => `@${o}`).join(", ")}.`,
```
0x4007, line 100: I wonder if its redundant to tag the assignees and then remove them. We can leave it for now but just a thought to consider.

0x4007 (approved):
QA looks good

src/helpers/remind-and-remove.ts
gentlementlegen: From what I've seen you get no notification when un-assigned, so I think this might be useful?

src/worker.ts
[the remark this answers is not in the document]
gentlementlegen: This applies to all the plugins, I think it would be nice to have it in the news docs that are being written right now which is why I left a comment there: https://github.com/ubiquity/business-development/pull/94

whilefoo:
> The deadline takes the priority into account, which means that 7 days will become 2.4 days on a priority 3 task, for example. Yet, the message told that you would get a reminder after 7 days. This is properly displayed here: [Meniole#6 (comment)](https://github.com/Meniole/daemon-disqualifier/issues/6#issuecomment-2519463850)

I see, however I didn't get any reminders, is that because the bot doesn't run as a cron but only on events?

gentlementlegen:
@whilefoo It runs only on events indeed. Another key behavior is that if you don't link a pull-request before the first reminder is sent, you instantly get disqualified, which I believe is what happened to you. In case a PR is linked, you will get the reminder in the pull-request itself.

PR #61, opened 2024-12-09
fix: pull-request state adjustments

### LINKED ISSUE #59: Pull Request State Adjustments
- When a follow up occurs, the pull request should be turned into a draft.
- If it is already a draft, only leave the follow up message.[^01^] 
- When a disqualification occurs, the pull request should be closed. 

Context: https://github.com/ubiquity-os/plugins-wishlist/issues/60#issuecomment-2516409360

Rationale: when a contributor converts from draft to ready, then that signals it is ready for review. 

[^01^]: ⚠ 52% possible duplicate - [Reviewer Follow Ups](https://www.github.com/ubiquity-os-marketplace/daemon-disqualifier/issues/49#49)

gentlementlegen:
Resolves #59
QA: https://github.com/Meniole/daemon-disqualifier/pull/8#issuecomment-2529008128


package.json
```diff
@@ -32,13 +32,12 @@
   "dependencies": {
     "@octokit/graphql-schema": "^15.25.0",
     "@octokit/rest": "^21.0.2",
-    "@sinclair/typebox": "0.34.3",
-    "@ubiquity-os/plugin-sdk": "^1.1.0",
+    "@sinclair/typebox": "0.34.11",
```
whilefoo, line 35: I think we should leave it at `0.34.3` same as in SDK because I noticed even between patch releases there are Typescript errors when passing the schema to `createPlugin`

src/worker.ts
```diff
@@ -0,0 +1,26 @@
...
+  (context) => {
+    return run(context);
+  },
+  manifest as Manifest,
+  {
+    envSchema: envSchema,
+    settingsSchema: pluginSettingsSchema,
+    logLevel: process.env.LOG_LEVEL || LOG_LEVEL.INFO,
+    postCommentOnError: false,
+    ...(process.env.KERNEL_PUBLIC_KEY && { kernelPublicKey: process.env.KERNEL_PUBLIC_KEY }),
```
whilefoo, line 18: why `process.env`?

whilefoo (approved):

src/worker.ts
gentlementlegen: This works because it is ran with `bun server`

package.json
gentlementlegen: Reverted the version.

whilefoo: Ideally I'd remove the package from plugin's dependencies and somehow use SDK's version, that way we don't need to change all plugins when we change the version in the SDK

src/worker.ts
whilefoo: ok so because this is only used for testing locally?

package.json
gentlementlegen: Maybe we could list it in the `peerDepencies` so the same version shall be required in the plugin when installing the sdk?

PR #66, opened 2024-12-14
Merge development into main

gentlementlegen:


0x4007:
The amount of deployments here looks like a misconfiguration

gentlementlegen:
@0x4007 It seems that the runs of the plugin are somehow considered as deployments so the large amount makes sense, but I don't understand what is the criteria to be shown as "deployment".

PR #69, opened 2025-01-07
fix: the warning for the disqualification is not posted on tasks with no price label

### LINKED ISSUE #63: Do not run for non-funded task
> > [!IMPORTANT]
> > - Be sure to link a pull-request before the first reminder to avoid disqualification.
> > - Reminders will be sent every `7 days` if there is no activity.
> > - Assignees will be disqualified after `14 days` of inactivity.
> 
> 
> 

We should only do disqualification when there are labels and it's funded  

Also maybe you should put some effort into writing proper specs, at least until we get issue matching and task recommendation to work.

_Originally posted by @0x4007 in https://github.com/ubiquity-os-marketplace/text-conversation-rewards/issues/212#issuecomment-2537926078_

The disqualifier should not run on task that are not funded (e.g. have no price label).
            
## What happened

The `daemon-disqualifier` ran and sent a message on an issue that was not priced, thus not subject to disqualification (no deadline needed).

## What was expected

`daemon-disqualifier` should run on priced tasks, and ignored unpriced tasks.

## How to reproduce

- create a task without any price label
- assign a user to that task
- notice the notification message being sent by `daemon-disqualifier`

gentlementlegen:
Resolves #63
QA: https://github.com/Meniole/daemon-disqualifier/issues/10#issuecomment-2575151732


src/handlers/watch-user-activity.ts
```diff
@@ -48,8 +48,16 @@ export async function watchUserActivity(context: ContextPlugin) {
...
+ * We ignore the issue if:
+ * - draft
+ * - pull request
+ * - locked
+ * - not in "open" state
+ * - not priced (no price label found)
+ */
 function shouldIgnoreIssue(issue: IssueType) {
-  return issue.draft || issue.pull_request || issue.locked || issue.state !== "open";
+  return !!issue.draft || !!issue.pull_request || issue.locked || issue.state !== "open" || parsePriceLabel(issue.labels) === null;
```
whilefoo, line 60: why is `!!` needed if `draft` is a boolean?

gentlementlegen: Very interesting question, because I get a warning when I apply `!!` to `locked` which is a boolean, but no warning on `draft`, I suppose because it is `boolean | undefined`, but even without `!!` it gets coerced to a boolean. I had added that because this function returned `boolean | {}` due to `pull_request`. Will remove `!!`.

PR #73, opened 2025-01-22
feat: cron job

### LINKED ISSUE #68: Use the proper configuration (repo scoped)
@0x4007 So the configuration is correct and does not require a pull-request as you can see [here](https://github.com/ubiquity/business-development/blob/development/.github/.ubiquity-os.config.yml#L370). However during the run that value was `true` as you can see [here](https://github.com/ubiquity-os-marketplace/daemon-disqualifier/actions/runs/12478009115/job/34824958340#step:3:26). This is probably due to the fact that the run was triggered from another repo where that value is `true`. Not sure how we should handle this scenario.

_Originally posted by @gentlementlegen in https://github.com/ubiquity/business-development/issues/103#issuecomment-2560735421_

gentlementlegen:
Resolves #68

## What are the changes

In this PR, I changed the behavior of the plugin to only update reminders on the repository where the reminders need to be sent. This ensures that the proper configuration is used, unlike now where we use any configuration where the event was triggered and apply it organization wide.

## How does it work

It was actually tricky to make it update only the concerned repos, according to a CRON. The flow is the following:

- when the workflow is triggered by events, we **ONLY** update the repository, with the repository configuration
- when this is triggered by `issues.assigned`, the posted comment for reminder is stored in a json file
- the CRON script runs once a day, and its only action is to read all the items stored in the db.json. For each of these, it will update the comment, happening an invisible content to it
- this in turns wakes up the kernel that will dispatch a `issue_comment.edited` event, so the `compute.yml` will run in the proper repository
- when a user is disqualified, the item in the db.json is removed
- if no more items are left in the `db.json` the `cron.yml` gets disabled
- if a new user gets assigned, then the `cron.yml` gets enabled again

## Caveats

- CRON can only run in the default branch. Thus, only the code from the default branch can be run
- likewise, the `db.json` will be only in the default branch, otherwise the cron can't read the data. This means that if we use the plugin from another branch (for example `ubiquity-os-marketplace@main`, all the data will be stored in the development branch.
- when multiple runs happen simultaneously, data is lost in the db (gets automatically merged by GitHub, [example here](https://github.com/Meniole/daemon-disqualifier/commit/5cbd97f785c874a078947b17ea1bfe67cc25403c))

I really think that this json storage is not reliable at all, and will lead to issues later on.


0x4007:
Seems like it might be tedious to show QA but are you able to show any

gentlementlegen:
There is of course, but I don't know if you can make much sense of it.

QA:
- [disqualified after inactivitiy](https://github.com/Meniole/daemon-disqualifier/issues/11#issuecomment-2612147579) (you can see that the start comment got modified by the bot)
- [cron runs, and disabled cron after no activity left to watch](https://github.com/Meniole/daemon-disqualifier/actions/workflows/cron.yml)
- [disqualified on another repo simultaneously](https://github.com/Meniole/text-conversation-rewards/issues/47#issuecomment-2611828935)

0x4007 (approved):
I guess let's merge and try in production I think this one is difficult to QA

.github/workflows/compute.yml
```diff
@@ -42,3 +46,22 @@ jobs:
 
       - name: Watch Activity
         uses: ./
+
+      - name: Get GitHub App token
+        if: env.APP_ID != '' && env.APP_PRIVATE_KEY != ''
+        uses: tibdex/github-app-token@v1.7.0
```
whilefoo, line 52: don't we use [actions/create-github-app-token](https://github.com/actions/create-github-app-token) now?

src/handlers/watch-user-activity.ts
```diff
@@ -91,6 +106,14 @@ async function updateReminders(context: ContextPlugin, repo: ListForOrg["data"][
...
       } else {
         logger.info(`Skipping issue ${issue.html_url} because no user is assigned.`);
+        const { owner, repo } = parseIssueUrl(issue.html_url);
+        const key = `${owner}/${repo}`;
+        if (db.data[key]) {
+          await db.update((data) => {
+            data[key] = data[key].filter((o) => o.issueNumber !== issue.number);
+            return data;
+          });
+        }
```
whilefoo, line 116: isn't this basically the same as calling `removeEntryFromDatabase`?

.github/workflows/compute.yml
gentlementlegen: Right, will also update the other workflow.

whilefoo (approved):

PR #79, opened 2025-02-06
feat: top ups

### LINKED ISSUE #51: Time "Top Up" Limits
A quick idea that can use improvement:

Assignee can restart the disqualification timer by continuing to try. There's many clear examples of the task not being the right fit for the assignee and they try i.e. for months on projects that should be doable in a day etc. 

We could consider doing limits on time limit "top ups." So for example:

Priority 1 - 5 top ups
Priority 2 - 4 top ups 
Priority 3 - 3 top ups 
Priority 4 - 2 top ups 
Priority 5 - 1 top up. 

I suspect that priority 5 tasks will border on eliminating most even capable assignees but once the task is complete even if just slightly after deadline we can always retroactively assign credit. 

---

- We should be able tot toggle this behavior in the config. 
- if it is enabled the message should clearly explain the amount of remaining top ups. 

I foresee a bit of complexity around the logic related to dealing with the "time limits" for follow up/disqualify and the top up counter. 
- If the limits are 7/14 as they are now, then the first top up should occur after 14 days on a priority 1 normal project. 
- priority 5 top up would be after `14 / 5 = 2.8` days.

gentlementlegen:
Resolves #51
QA: https://github.com/Meniole/daemon-disqualifier/issues/14#issuecomment-2652947298


src/run.ts
```diff
@@ -1,7 +1,53 @@
...
+  });
+
+  if (!data.length) {
+    logger.debug("No labels have been found, won't populate top ups.", { data });
+    return;
+  }
+
+  const priorityLabels = data
+    .map((label) => {
+      const match = label.name.match(/^Priority:\s*(\d+).*$/i);
```
0x4007, line 26: Ideally we should parse the priority labels from the plugin config because they are intended to be customizable to adhere to each org's preferences.

src/helpers/remind-and-remove.ts
```diff
@@ -40,12 +42,14 @@ async function remindAssignees(context: ContextPlugin, issue: ListIssueForRepo)
...
   }
+
   const logins = issue.assignees
     .map((o) => o?.login)
     .filter((o) => !!o)
     .join(", @");
+  const topUps = await getTopUpsRemaining(context);
 
-  const logMessage = logger.info(`@${logins}, this task has been idle for a while. Please provide an update.\n\n`, {
+  const logMessage = logger.info(`@${logins}, this task has been idle for a while. Please provide an update.\n\n<h5>Top ups remaining: ${topUps}</h5>`, {
```
0x4007, line 52: https://github.com/ubiquity-os-marketplace/daemon-disqualifier/issues/51#issuecomment-2647547827

src/helpers/remind-and-remove.ts
```diff
@@ -40,14 +42,19 @@ async function remindAssignees(context: ContextPlugin, issue: ListIssueForRepo)
...
     .map((o) => o?.login)
     .filter((o) => !!o)
     .join(", @");
+  const { remainingTopUps, topUpLimit } = await getTopUpsRemaining(context);
 
-  const logMessage = logger.info(`@${logins}, this task has been idle for a while. Please provide an update.\n\n`, {
-    taskAssignees: issue.assignees.map((o) => o?.id),
-  });
+  const logMessage = logger.info(
+    `@${logins}, this task has been idle for a while. Please provide an update.\n\n<h5>You used ${topUpLimit - remainingTopUps} / ${topUpLimit} top ups</h5>`,
```
0x4007, line 53:
```suggestion
    `@${logins}, this task has been idle for a while. Please provide an update.\n\n<h5>You used `${topUpLimit - remainingTopUps}` of `${topUpLimit}` top ups</h5>`,
```

src/run.ts
gentlementlegen: Right but at the moment there is not way to read other plugins configuration, and even if we could I could not guess that this section is the `Priority`. And this part of the code is just a fallback if nothing is set in the configuration, because I consider the default as `Priority`.

0x4007: Anything is possible. Just parse the yml config looking for daemon pricing

gentlementlegen: Yes but how do I guess which parameter of daemon pricing are meant to be the priority labels? The problem is the same.

gentlementlegen:
I need to fix the tests because the whole logic of reminders changed, but the code can be reviewed.

0x4007:
There's no top up messages so this doesn't achieve the spec.

gentlementlegen:
@0x4007 what do you mean by "there is no top-up message"? It gives you you current top-ups, what other message is needed?

0x4007:
I guess your examples only say "disqualified cause no more top ups remaining"

Can you show an example with the counter like we discussed 

"You used 3 of 5 top ups!" Etc

gentlementlegen:
This is because you are checking the issue, you should check the pull-request attached to that issue: https://github.com/Meniole/daemon-disqualifier/pull/15

I linked the issue so we can see both the top up exhausted message from the PR and the issue.

0x4007:
Can you show a full example because the one you just linked only says 1 of 2 both times then it's closed which is unexpected. 

It should be 1 of 2 

2 of 2 

Disqualified.

gentlementlegen:
It can be many times 1/2 and 2/2 depending on the frequency of the reminders. I'll redo a clear test then.

0x4007:
/ask rewrite the message to be clear and concise given the context of this feature

0x4007:
/ask rewrite the message to be clear and concise given the context of this feature

gentlementlegen:
@0x4007 Updated QA: https://github.com/Meniole/daemon-disqualifier/pull/17#issuecomment-2661167201

0x4007:
It would seem more like a human manager if it didn't constantly remind of the top up state. 

Ideally we should only remark on the top up state on state change. 

Alternatively we can reword the entire comment to make it seem like it's a foot note/subtitle or something.

gentlementlegen:
I asked you if it was fine to have numerous reminders on each top up stage and you had confirmed it was okay which is why I did it this way. If we do one reminder per top up, I think we should totally remove the warning threshold and disqualification because it would be pointless (both cannot coexist).

Consider the following scenario:
- warning 5 minutes
- disqualification 10 minutes
- time: 1 week
- priority: 5

In this PR (the way it is right now), you would get one reminder every 5 minutes, and eventually disqualified after 10 minutes of total inactivity.

If we do one reminder per top up, it would mean one reminder every 1.4 days, making the warning threshold and disqualification threshold obsolete.

0x4007:
Perhaps it makes more sense to focus on making it better UX for expected follow up times instead of optimizing it for a testing edge case. Nobody is going to set a 10 minute threshold in production.

gentlementlegen:
I exxaggerated so you understand where the problem is. Same problem goes with our current configuration of 7 days warning / 14 days disqualification. With a < 15 minutes task priority 5, you would not get the top ups saying 1/2 and 2/2 because this would be impossible.

0x4007:
We need first to use the time label as a given amount of time and then use our disqualifier configured timer 

So 15 minutes would be the time label and then after that we hit the first top up

Then after 7 days or whatever the second top up.

gentlementlegen:
@0x4007 But the warning threshold would be hit several times so several warning will be issues with the same top-up amount in between.

0x4007:
> @0x4007 But the warning threshold would be hit several times so several warning will be issues with the same top-up amount in between.

I don't understand but I imagine that the previous comments can be reviewed for a string match to determine when to share the update about "state change"

> This is not at all the issue specification. It states:

It's under remarks and emphasizes multiple times in our DMs as a "concern"

gentlementlegen:
What I am saying is given what you describe, both systems cannot coexist. Either we use top-ups, either we use disqualification threshold. Having both does not make sense if we want only to remind on top up consumed.

0x4007:
Sure. Bonus points if you can consolidate into a single comment next time cron runs

gentlementlegen:
New QA: https://github.com/Meniole/daemon-disqualifier/pull/17#issuecomment-2669047094

0x4007:
I left comments. New QA I wouldn't have expected it to close it until disqualification. If it is pull related perhaps it would be more clear to mention that no pull is linked and in this repository a pull must be linked for deadline extensions. 

For the second link I think that making the numbers bold as well looks better. I left an example there.

gentlementlegen:
I commented there: it disqualified because I do not show any activity during the last 15 minutes, which was above the disqualification threshold, so I got disqualified regardless of top ups.

src/helpers/task-metadata.ts
```diff
@@ -47,6 +35,29 @@ export async function getTaskAssignmentDetails(
...
+ *
+ * It does not care about previous updates, comments or other events that might have happened on the issue.
+ *
+ * It returns who is assigned and the initial calculated disqualification threshold (start + time label duration).
+ */
+export async function getTaskAssignmentDetails(
+  context: ContextPlugin,
+  repo: ContextPlugin["payload"]["repository"],
+  issue: ListIssueForRepo
+): Promise<{ startPlusLabelDuration: string; taskAssignees: number[] } | false> {
```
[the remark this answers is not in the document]
gentlementlegen: It is set in the metadata as informational but doesn't appear in any logic or calculations. I can remove it if it is not deemed useful.

src/helpers/remind-and-remove.ts
```diff
@@ -22,41 +24,129 @@ export async function remindAssigneesForIssue(context: ContextPlugin, issue: Lis
...
+  pr?: {
+    prOwner: string;
+    prRepo: string;
+    prNumber: number;
+  };
+}
 
-  if (!issue?.assignees?.length) {
-    logger.error(`Missing Assignees from ${issue.html_url}`);
+async function shouldDisplayTopUpsReminder(context: ContextPlugin, args: Args) {
```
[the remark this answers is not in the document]
gentlementlegen: I renamed them if that helps to make it clearer. Basically it just that sometimes we need to check inside the pull-request(s) or the main issue, if there is no linked PR.

src/helpers/task-update.ts
```diff
@@ -7,17 +7,44 @@ import { collectLinkedPullRequests } from "./collect-linked-pulls";
...
-import { closeLinkedPullRequests, remindAssigneesForIssue, unassignUserFromIssue } from "./remind-and-remove";
+import { closeLinkedPullRequests, remindAssignees, remindAssigneesForIssue, unassignUserFromIssue } from "./remind-and-remove";
 import { getCommentsFromMetadata } from "./structured-metadata";
 import { getTaskAssignmentDetails, parsePriorityLabel } from "./task-metadata";
 
 function getMostRecentActivityDate(assignedEventDate: DateTime, activityEventDate?: DateTime): DateTime {
   return activityEventDate && activityEventDate > assignedEventDate ? activityEventDate : assignedEventDate;
 }
 
+export async function getAssignedEvent(context: ContextPlugin, repo: ContextPlugin["payload"]["repository"], issue: ListIssueForRepo) {
```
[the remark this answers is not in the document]
gentlementlegen: It seems to consider the metadata when getting assignment details, the other function does not. To be honest I just split this to lower the complexity of the other function, I didn't change any logic.

gentlementlegen:
@0x4007 I renamed code and configuration occurrences to match your comment: https://github.com/Meniole/daemon-disqualifier/pull/17#issuecomment-2676522083

src/run.ts
0x4007: Since you added the priority levels to the config here, why don't you work with those?

0x4007 (commented):
https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/79#discussion_r1966956414

src/run.ts
gentlementlegen: I am using these, it's just that if they are omitted they automatically get populated, to make it more user friendly.

0x4007 (approved):

src/run.ts
gentlementlegen: I think we should enforce some patterns, because if they decide to do "Priority Low", "Priority Medium" etc. we can't determine what priority is higher than the other (except if we use an LLM or rely on the order in the array). This can be discussed in another task, but so far our system relies a lot on names (we expect "Price:" or "Time:")

src/helpers/remind-and-remove.ts
```diff
@@ -22,41 +34,135 @@ export async function remindAssigneesForIssue(context: ContextPlugin, issue: Lis
...
+    issue,
+  }: {
+    reminderContent: string;
+    extensionsLimit: number;
+    remainingExtensions: number;
+    owner: string;
+    repo: string;
+    issue_number: number;
+    issue: ListIssueForRepo;
+  }
```
whilefoo, line 120: why are there so many defined properties if only 4 of them are used?

src/run.ts
```diff
@@ -1,7 +1,53 @@
 import { watchUserActivity } from "./handlers/watch-user-activity";
 import { ContextPlugin } from "./types/plugin-input";
 
+async function populateTopUpThresholds(context: ContextPlugin) {
```
whilefoo, line 4: you forgot to rename to deadline extensions

src/types/plugin-input.ts
```diff
@@ -91,6 +77,13 @@ export const pluginSettingsSchema = T.Object(
       description: "Whether a pull request is required for the given issue on disqualify",
       examples: ["true", "false"],
     }),
+    availableDeadlineExtensions: T.Object(
```
whilefoo, line 80: add description for plugin installer

src/helpers/task-update.ts
```diff
@@ -81,20 +99,24 @@ export async function updateTaskReminder(context: ContextPlugin, repo: ContextPl
     mostRecentActivityDate: mostRecentActivityDate.toLocaleString(DateTime.DATETIME_MED),
   });
 
-  const disqualificationTimeDifference = disqualification - warning;
+  const disqualificationTimeDifference = negligenceThreshold - followUpInterval;
 
   if (lastReminderComment) {
     const lastReminderTime = DateTime.fromISO(lastReminderComment.created_at);
     mostRecentActivityDate = lastReminderTime > mostRecentActivityDate ? lastReminderTime : mostRecentActivityDate;
```
whilefoo, line 106: I know this is old code but I'm trying to understand the whole logic, won't this be always `lastReminderTime` because last reminder is always newer than `mostRecentActivityDate` or am I missing something?

src/helpers/task-metadata.ts
```diff
@@ -8,37 +8,25 @@ type IssueLabel = Partial<Omit<RestEndpointMethodTypes["issues"]["listLabelsForR
...
     .sort((a, b) => DateTime.fromISO(b.created_at).toMillis() - DateTime.fromISO(a.created_at).toMillis());
 
-  const latestUserAssignment = assignedEvents.find((o) => o.actor?.type === "User");
-  const latestBotAssignment = assignedEvents.find((o) => o.actor?.type === "Bot");
-
+  const latestUserAssignment = assignedEvents.find((o) => ("assigner" in o ? o.assigner?.type === "User" : o.actor?.type === "User"));
+  const latestBotAssignment = assignedEvents.find((o) => ("assigner" in o ? o.assigner?.type === "Bot" : o.actor?.type === "Bot"));
   let mostRecentAssignmentEvent = latestUserAssignment || latestBotAssignment;
 
   if (latestUserAssignment && latestBotAssignment && DateTime.fromISO(latestUserAssignment.created_at) > DateTime.fromISO(latestBotAssignment.created_at)) {
```
whilefoo, line 32: looking at this again and it doesn't make sense, first `mostRecentAssignmentEvent` gets assigned to whichever is truthy and then it's checked if both are present and if user assignment is newer than bot then user assignment gets assigned to `mostRecentAssignmentEvent` but in else statement it's assigned to bot assignment even if it might actually be `null`

src/helpers/remind-and-remove.ts
```diff
@@ -22,41 +34,135 @@ export async function remindAssigneesForIssue(context: ContextPlugin, issue: Lis
...
+  const repo = issueAndPrTargets.pr?.prRepo ?? context.payload.repository.name;
+
+  if (!owner) {
+    logger.error("No owner was found in the payload, won't display top-up value");
+    return false;
+  }
+
+  const regex = new RegExp(/"remainingTopUps": (\d+)/, "i");
+
+  const lastTopUpValue = (
```
whilefoo, line 69: you forgot to rename

src/helpers/task-metadata.ts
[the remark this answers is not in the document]
whilefoo: if it gets logged that's fine, but if it's not used then we should remove it

src/helpers/task-update.ts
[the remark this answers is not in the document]
whilefoo: I'm not sure if the metadata affects anything because it's just an extra check that assignment events assignees are really the ones that started the task so maybe we should use the one with metadata?

src/helpers/task-update.ts
gentlementlegen: `lastReminderTime` should be the time when the last reminder comment was posted by the bot. `mostRecentActivityTime` should be the last time the assigned user had some activity (push, comment), so not necessarily.

src/helpers/task-metadata.ts
gentlementlegen: Added a check to avoid assigning a null value if it is null: https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/79/commits/7bda7cb66a05a1e5a78cc12aeb6e4f014db2954e

gentlementlegen:
New QA after code changes: https://github.com/Meniole/daemon-disqualifier/issues/20

0x4007 (approved):
Just need to change the message and we can merge this https://github.com/Meniole/daemon-disqualifier/issues/20#issuecomment-2685408080

src/helpers/task-update.ts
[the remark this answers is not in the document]
gentlementlegen: https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/79/commits/e84e9955a75f2f38b9bf9182ebc173438a252d69

src/helpers/task-metadata.ts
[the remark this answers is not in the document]
gentlementlegen: https://github.com/ubiquity-os-marketplace/daemon-disqualifier/pull/79/commits/e84e9955a75f2f38b9bf9182ebc173438a252d69

src/helpers/task-update.ts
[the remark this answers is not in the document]
gentlementlegen: Bit confused, maybe you meant `the one without metadata`?

src/helpers/task-update.ts
whilefoo: if you look at the code (line 71) you can see that it only checks for reminders that happened after most recent activity

src/helpers/task-update.ts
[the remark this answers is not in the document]
whilefoo: Why? The one with metadata just has an extra check. Honestly just choose one, I don't think it makes much difference in practice :D

src/helpers/task-update.ts
gentlementlegen: There is nothing for me on line 71, but I did see that the filter removes events that are before. sorry for tiptoeing, this code gave me quite a few headaches already.

src/helpers/task-update.ts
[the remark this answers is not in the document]
gentlementlegen: I unified the two functions, but it does make a difference because to fetch the metadata the whole issue has to be passed in (it checks for assignees, labels, and html_url) which we don't have when we check the linked pull-requests. To avoid re-fetching every linked pull-request, I made an option to pass in a number or an Issue. If a number is passed in then it won't check the metadata. I am also updating the tests that broke.

whilefoo (approved):

PR #88, opened 2025-04-06
fix: extensions

### LINKED ISSUE #86: Fix unexpected disqualifications
There have been a couple of instances where the assignee was unassigned due to supposedly using all available deadline extensions but should not have been.

First case is where PR was approved but the assignee was unassigned anyway
Examples: 
https://github.com/ubiquity-os-marketplace/daemon-pull-review/issues/23#issuecomment-2739822030
https://github.com/ubiquity-os-marketplace/daemon-pricing/issues/64#issuecomment-2707859584
The problem in the code is that the PR is checked for approvals only if the reminder has been sent, it should be checked even if there weren't any reminders.

Second case is where the assignee got unassigned but no reminders have been sent.
I can't find any examples but I'm pretty sure there was such a case.
The solution is to first send a reminder and then disqualify the user.

gentlementlegen:
Resolves #86

This PR contains a few changes:
- on user assigned event, we only display the reminder and do not run the deadline checks, because this is the cron's job (avoids double posts)
- some functions were not targeting the proper PR / issue when checking for updates, which should be solved
- added LOG_LEVEL env var, linguist generated for the db not to count in rewards
- the structured metadata now outputs the run url
- updated plugin inputs to actually reflect the events handled by the plugin
- changed the run to be synchronous, because there are a lot of issues with the db being overwritten by all the runs going on at the same time
- we do not display deadline quota if the pull-request is approved and the reminder is to nudge the reviewers
- we were not handling the case where a user got assigned through the PR and not through the issue (e.g. assigning itself by opening a pull-request), a fallback is now available
- logs have generally been improved for debug

QA: lately, `ubiquity-os-marketplace` has been running on this branch with the changes.


src/helpers/structured-metadata.ts
```diff
@@ -13,7 +14,8 @@ export function createStructuredMetadata(className: string, logReturn: LogReturn
   const jsonPretty = JSON.stringify(metadata, null, 2);
   const stackLine = new Error().stack?.split("\n")[2] ?? "";
   const caller = stackLine.match(/at (\S+)/)?.[1] ?? "";
-  const ubiquityMetadataHeader = `<!-- ${HEADER_NAME} - ${className} - ${caller} - ${metadata?.revision}`;
+  const runUrl = github?.context ? `${github.context.payload.repository?.html_url}/actions/runs/${github.context.runId}` : "http://localhost";
+  const ubiquityMetadataHeader = `<!-- ${HEADER_NAME} - ${className} - ${caller} - ${metadata?.revision} - ${runUrl}`;
```
whilefoo, line 18: don't we have this metadata creation in the sdk?

src/helpers/task-update.ts
```diff
@@ -78,18 +78,29 @@ export async function updateTaskReminder(context: ContextPlugin, repo: ContextPl
     if (await areLinkedPullRequestsApproved(context, issue)) {
       if (context.config.followUpInterval > 0) {
         // If the issue was approved but is not merged yet, nudge the assignee
```
whilefoo, line 80: why is this not checked only if there is `lastReminderComment`?
if there is no reminder it will call `remindAssigneesForIssue` which will unassign if all extensions are used even though the PR was approved

src/handlers/watch-user-activity.ts
```diff
@@ -42,14 +47,21 @@ export async function watchUserActivity(context: ContextPlugin) {
...
+    const repo = context.payload.repository;
+    logger.debug(`> Watching user activity for repo: ${repo.name} (${repo.html_url})`);
+    await updateReminders(context, repo);
+    await updateCronState(context);
 
-  return { message: "OK" };
+    return { message: "OK" };
+  } else if (!isIssueComment(context)) {
+    return { message: logger.warn("The comment is not related to any daemon-disqualifier comment edit.").logMessage.raw };
+  }
```
whilefoo, line 63: this log message is confusing - if the `if` checks that it's not an issue comment then why does the message say the comment is not related

src/helpers/deadline-extensions.ts
```diff
@@ -1,34 +1,35 @@
...
 
-  const priorityList = Object.keys(context.config.availableDeadlineExtensions.amounts);
-  const priorityLabel = context.payload.issue.labels?.find((label) => priorityList.includes(label.name));
+  const priorityList = Object.keys(config.availableDeadlineExtensions.amounts);
+  const priorityLabel = issue.labels?.find((label) => typeof label !== "string" && label.name && priorityList.includes(label.name));
 
-  if (!priorityLabel) {
+  if (!priorityLabel || typeof priorityLabel === "string" || !priorityLabel.name) {
     return defaultExtensions;
   }
```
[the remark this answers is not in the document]
gentlementlegen: Yes it is, that's to avoid a TS error line 27 when doing `priorityLabel.name`

src/helpers/structured-metadata.ts
gentlementlegen: We do but it is not exposed
https://github.com/ubiquity-os/plugin-sdk/blob/da526ffb2dbb4b008af87f60e7755c4f12922af2/src/comment.ts#L157-L168

the problem is that in this plugin we sometimes post to pull-requests, to issues, and even to multiple PRs which the SDK doesn't provide a way to do easily

src/helpers/deadline-extensions.ts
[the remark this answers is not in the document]
gentlementlegen: Yes it is, that's to avoid a TS error line 27 when doing priorityLabel.name

src/helpers/task-update.ts
gentlementlegen: You are right, I did not think about that scenario. Will put this outside of the conditions and return early in that case.

whilefoo (approved):

PR #100, opened 2025-08-10
chore(kv): migrate to kv-based database handling

### LINKED ISSUE #99: Migrate from `db.json` to KV storage
Like we previously did for https://github.com/ubiquity-os-marketplace/daemon-merging/issues/51 we would benefit from moving to storage from `db.json` to Deno KV, which would make CRON updates from this plugin much smoother.

gentlementlegen:
Resolves #99

### DIFF
--- a/src/adapters/kv-database-handler.ts
+++ b/src/adapters/kv-database-handler.ts
@@ -1,0 +1,76 @@
+import { parseIssueUrl } from "../helpers/github-url";
+
+export const KV_PREFIX = "cron";
+
+export class KvDatabaseHandler {
+  private _kv: Deno.Kv;
+
+  constructor(kv: Deno.Kv) {
+    this._kv = kv;
+  }
+
+  async getIssueNumbers(owner: string, repo: string): Promise<number[]> {
+    const key = [KV_PREFIX, owner, repo];
+    const result = await this._kv.get<number[]>(key);
+    return result.value || [];
+  }
+
+  async addIssue(url: string): Promise<void> {
+    const { owner, repo, issue_number } = parseIssueUrl(url);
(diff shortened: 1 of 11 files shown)


src/adapters/kv-database-handler.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private async _getIssueEntries(owner: string, repo: string): Promise<IssueEntry[]> {
+    const key = [KV_PREFIX, owner, repo];
+    const result = await this._kv.get<IssueEntry[] | number[]>(key);
+    if (!result.value) return [];
+    const value = result.value;
+    if (Array.isArray(value) && value.length > 0 && typeof value[0] === "object" && value[0] !== null && "issueNumber" in value[0]) {
+      return value as IssueEntry[];
+    }
+    return (value as number[]).map((issueNumber) => ({ issueNumber, commentId: 0 }));
```
whilefoo, line 25: Why does it have two ways of storing the issues?

tests/main.test.ts
```diff
@@ -18,8 +18,6 @@ import { server } from "./__mocks__/node";
 import cfg from "./__mocks__/results/valid-configuration.json";
 import { botReminderComment, getIssueHtmlUrl, STRINGS } from "./__mocks__/strings";
 
-dotenv.config();
-
 beforeAll(() => {
   server.listen();
 });
```
whilefoo, line 23: You might want to create a new instance of the mock server (and db) for each test file for the future if multiple test files will use the server as it won't work with shared instance.

gentlementlegen: You're right I didn't notice because only this test suite creates the server, will change it

PR #121, opened 2026-02-09
feat: run cron via repo octokit and remove commentId tracking

### LINKED ISSUE #120: Improve comment update system
Currently, when a CRON update is triggered, the comment posted by the bot about disqualification is edited to trigger the run. It is not that big of a deal, because this is a bot comment and does not go in the vectors nor has much relevance to the scoring and results. However, this means the comment gets longer and longer each update, which does not seem right.

There could be two approaches to this problem

## Update the time on the hidden comment instead of append

This would make sure only one comment is added to the start stop, where we can just update the time. The approach is simple and requires little changes.

## Directly create an Octokit instance for that repo and start a run entirely

This would avoid comment and reduce triggered events, but requires lots of changes like pulling the proper configuration with the Plugin SDK, mocking the event payload to trigger a run, and is more risky overall also regarding permissions.

gentlementlegen:
## Summary
- replace comment-edit CRON trigger with direct synthetic `issue_comment.edited` runs using installation-scoped Octokit clients
- keep KV entries as object arrays and remove runtime `commentId` dependency (`{ issueNumber }` only)
- keep one-off migration/seed scripts local-only (not part of this PR)
- update docs, cron tests, and KV compatibility tests

## Testing
- bunx tsc --noEmit
- bun test --preload=./tests/setup.ts tests/cron.test.ts tests/assign.test.ts tests/kv-database-handler.test.ts

Closes #120

QA: https://github.com/Meniole/daemon-disqualifier/actions/runs/21814544411/job/62933370907
QA2: https://github.com/Meniole/daemon-disqualifier/actions/runs/21816207874/job/62938296509

Removed the issue edited event, so we should have less noise.

After closing the task, I'll need to run a script to properly update KV content.


src/cron/runner.ts
```diff
@@ -40,13 +72,92 @@ async function enforceRateLimit(): Promise<void> {
...
+  });
+}
+
+async function resolveRepoConfig(octokit: ContextPlugin["octokit"], owner: string, repo: string): Promise<PluginSettings | null> {
+  try {
+    const handler = new ConfigurationHandler(logger, octokit);
+    const parsedConfig = await handler.getSelfConfiguration(manifest, { owner, repo });
+    if (!parsedConfig) {
+      return null;
+    }
```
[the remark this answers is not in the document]
gentlementlegen: `getSelfConfiguration` will try to pull the repository configuration and then the organization config. If both are missing, it returns `null` and it should be considered that this repo should entirely be skipped.

src/helpers/structured-metadata.ts
```diff
@@ -33,8 +32,6 @@ export function createStructuredMetadata(className: string, logReturn: LogReturn
   return metadataSerialized;
```
0x4007, line 32: If we got rid of relying on the metadata in the comment, then perhaps we should remove all this code as well.

src/cron/runner.ts
```diff
@@ -40,13 +72,92 @@ async function enforceRateLimit(): Promise<void> {
   }
 }
 
+async function getInstallationOctokit(appOctokit: ContextPlugin["octokit"], owner: string, repo: string, appAuth: AppAuth) {
```
0x4007, line 75: Shouldn't this be in our SDK as a shared method? Pretty sure I commented on this in another pull request.

0x4007 (approved):

src/cron/runner.ts
gentlementlegen: It wouldn't be very useful as it would be a one liner, we use `customOctokit` either way, but here we have to pass different `appId` for each org / repo which is why this function exists.

src/helpers/structured-metadata.ts
gentlementlegen: Good point, let me see if we use it for something.

gentlementlegen: I checked current code paths and the metadata is still useful because:
- Follow-up comments are stamped with this metadata when posted, and later we query those stamped bot comments to detect whether a reminder already exists before sending another one.
- The reminder flow also reads metadata content (e.g., `remainingExtensions`) from prior bot comments to decide whether to show/update the extension reminder message.
Without this metadata, reminder state detection would lose its source of truth in comment history.

src/cron/runner.ts
```diff
@@ -40,13 +72,92 @@ async function enforceRateLimit(): Promise<void> {
...
+    command: null,
+    commentHandler: new CommentHandler(),
+    config: args.config,
+    env: {},
+    eventName: "issues.reopened",
+    logger,
+    octokit: args.octokit,
+    payload: {
+      action: "reopened",
+      installation: { id: 0 },
```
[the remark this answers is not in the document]
gentlementlegen: Aligned implementation with the PR description.

Updated the synthetic CRON context to use:
- `eventName: "issue_comment.edited"`
- `payload.action: "edited"`

Also updated the CRON test expectation accordingly.

PR #131, opened 2026-03-17
fix(cron): resolve plugin manifests from dist branches

### LINKED ISSUE #130: Cron should resolve plugin manifests from dist/* branches
## Problem

The cron workflow resolves plugin manifests for config-referenced marketplace actions from the source branch ref such as `development`, but those manifests now live on artifact branches like `dist/development`.

Reference run from March 17, 2026:
https://github.com/ubiquity-os-marketplace/daemon-disqualifier/actions/runs/23184991125/job/67366182478

In that run, cron requests remote manifests such as:

- `https://api.github.com/repos/ubiquity-os-marketplace/daemon-disqualifier/contents/manifest.json?ref=development`
- `https://api.github.com/repos/ubiquity-os-marketplace/command-config/contents/manifest.json?ref=development`

Those return 404s, while the corresponding artifact branches exist and contain `manifest.json`, for example `dist/development`.

## Expected behavior

Cron should resolve GitHub action manifests from `dist/<ref>` first, then fall back to `<ref>` when no artifact branch exists.

## Scope

- Keep the fix local to cron config resolution in this repository.
- Do not change cron workflow checkout or repository config file lookup.
- Cover both the main plugin and other config-resolved marketplace action plugins.

gentlementlegen:
Resolves #130

## Summary
- add a cron-specific configuration handler that resolves GitHub action manifests from `dist/<ref>` before falling back to the source ref
- keep the change local to cron config loading and preserve existing matching for `daemon-disqualifier@<ref>` config entries
- add cron tests for dist-first lookup, source-ref fallback, and non-rewritten manifest lookups

## Testing
- `bun test --preload=./tests/setup.ts`

### DIFF
--- a/src/cron/configuration.ts
+++ b/src/cron/configuration.ts
@@ -1,0 +1,46 @@
+import { Value } from "@sinclair/typebox/value";
+import { ConfigurationHandler, isGithubPlugin } from "@ubiquity-os/plugin-sdk/configuration";
+import manifest from "../../manifest.json" with { type: "json" };
+import { ContextPlugin, PluginSettings, pluginSettingsSchema } from "../types/plugin-input";
+
+type ManifestPlugin = Parameters<ConfigurationHandler["getManifest"]>[0];
+
+function getDistRef(ref: string) {
+  return ref.startsWith("dist/") ? ref : `dist/${ref}`;
+}
+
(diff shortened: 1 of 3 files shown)


gentlementlegen:
@codex review

